PR #23, opened 2022-03-25
Original benchmark version

Peter9192:
This branch contains a minimally adapted version of the original code that I was able to run. 

Don't merge this branch! We can use it for benchmarking later on.


Peter9192:
Closing as I'm not planning to benchmarking against this version anymore.

PR #24, opened 2022-03-25
Use Python 3, apply code formatters, add .gitignore and environment.yml

Peter9192:


Peter9192:
History got messed up, abandoning in favour of #25

Issue #13, opened 2022-03-18
Code formatting


Peter9192:
It can be nice to use a code formatter. Consistent formatting helps to make the code readable, helps to avoid discussions about formatting, and can make for clearer diffs. A popular formatter for Python code is [black](https://black.readthedocs.io/en/stable/).

Imme1992:
Do this when code is nearly finished, then apply code formatter again

Imme1992:
Add instructions to developer documentation

ruudvdent:
closed as it is indeed already part of issue #144

PR #25, opened 2022-03-25
Use Python 3, apply code formatters, add .gitignore and environment.yml

### LINKED ISSUE #7: Python 3
Currently, code is written for Python 2. It might be a good idea to update it to Python 3 for future compatibility. In any case, we should add an environment file that installs the correct version of python.

Peter9192:
* Align the code with #23 
* Add environment file
* Upgrade to python 3
* Apply code formatters (black and isort) for consistent formatting
* Add .gitignore (to help prevent accidentally committing non-source files)

closes #7
refs #13

Since this is the first time the code formatters are applied, it leads to a very large diff that is difficult to review. On the flipside, future diffs will be much cleaner. For reviewing this PR, it is easiest to view the commits one at a time (start [here](https://github.com/WAM2layers/WAM2layers/pull/25/commits/92b6584a616f3e2b57d7881c2045404529108f87) and click next). For the big formatting changes, you don't have to check every line, since there are no code changes in those commits, only formatting.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,152 @@
+# Byte-compiled / optimized / DLL files
+__pycache__/
+*.py[cod]
+*$py.class
+
+# C extensions
+*.so
+
+# Distribution / packaging
+.Python
+build/
+develop-eggs/
+dist/
+downloads/
+eggs/
+.eggs/
+lib/
+lib64/
+parts/
+sdist/
+var/
+wheels/
+share/python-wheels/
+*.egg-info/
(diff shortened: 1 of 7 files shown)


Peter9192:
Preprocessing 7 days with this branch took one hour with this branch on my laptop, that's at least a factor 2 speedup over the original (though my laptop is not the most reliable benchmarking machine).

Imme1992 (approved):
Yes improvements look good

PR #32, opened 2022-03-25
Simplify paths

### LINKED ISSUE #31: Simplify data loading in preproc
Many lines of code in the preprocessing deal with reading data. There's room for simplification there. For example, instead of 

```py
def data_path(yearnumber, month, a):
    var1 = ... +"var1" + ...
    var2 = ... +"var2" + ...
    ...
    return [var1, var2, var3, var4, ...]
```

you could pass in the var name itself like so:

```py
def data_path(yearnumber, month, a, var):
    return ... + var + ...
```

You can also consider loading the data (and combining the 'normal' and 'eoy' segments) in the same function, then you can simply do 

```py
q = load_data(yearnumber, month, a, "Q")
```

Making this simpler also helps to make the code easier to use with data that are stored following a different filenaming convention.

Peter9192:
work in progress

closes #31

### DIFF
--- a/Fluxes_and_States_mat.py
+++ b/Fluxes_and_States_mat.py
@@ -82,102 +82,83 @@ else:  # leap
     )
 
 # other scripts use exactly this sequence, do not change it unless you change it also in the scripts
-def data_path(yearnumber, month, a):
+def data_path(year, month, a, var)
 
-    q_f_data = os.path.join(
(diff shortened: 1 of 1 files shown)


Imme1992 (approved):

Peter9192 (commented):
Confirmed it still runs

PR #34, opened 2022-04-08
Refactor preprocessing

Peter9192:
work in progress


Peter9192:
At this point data loading takes up no more than a second, interpolation down to under 5 minutes. 
Step 7 seems to take long for the first step but then speeds up significantly for subsequent days.
```
Preprocessing data for 2002-01-01 00:00:00, begin_time=0, final_time=31
before interpolation loop 12:26:57.070026
after interpolation loop 12:31:51.213932
check calculation water vapor, this value should be zero: -1.0623724122638123e-12
Step 1, 2, 3 finished, elapsed time since start: 0:04:56.193826
Step 4 finished, elapsed time since start: 0:04:56.570826
...
Step 6b finished, elapsed time since start: 0:05:19.046509
...
Step 7 finished, elapsed time since start: 0:16:34.894763
... # saving
Runtime fluxes_and_storages for day=1, year=2002 is 0:18:04.227511
...
Runtime fluxes_and_storages for day=2, year=2002 is 0:09:06.416929
Runtime fluxes_and_storages for day=3, year=2002 is 0:07:03.621994
```

Peter9192:
Also added some quick verification figures to see if the code still produces the same results. There seem to be some differences, but the overall pattern is the same. Haven't looked into it enough to tell where they're coming from. Is it just numerical or did I introduce (or remove) a bug?

image

Peter9192:
Found 2 code changes that both have to do with the time bookkeeping. The data is a bit annoying in that it is monthly data but it doesn't start at 00:00, but either at 06:00 or 03:00 depending on whether the data is 6- or 3-hourly. 

The first change was that I accidentally treated U10 and V10 as 6-hourly variables instead of 3-hourly, so I got the wrong indices there.

The second one is a bit more tricky. The original code did something like this: 
```
times_3hourly = slice(begin_time*2+1, (begin_time*2+count_time*2+1), 2)  
# for day 1, this evaluated to [0, 2, 4, 6, 8], or to 
# [03:00, 09:00, 15:00, 21:00, 03:00]
```
Notice that this doesn't align with the 6-hourly data, and also that it crosses the midnight boundary. I think this might have been unintended in the original code. 

In refactoring I first changed it to `slice(0, 10, 2)` (https://github.com/WAM2layers/WAM2layers/pull/34/commits/3cd0c5d52365434134b3802dccfd382aa822a9dd), and later to `slice(1, 10, 2)` ( https://github.com/WAM2layers/WAM2layers/pull/34/commits/2e897832932c4193602cf2ab69f125d37a03dd93) to make sure the times between 3-hourly and 6-hourly variable would align. I think this correction is necessary, but for verification, I'll temporarily revert it.

_On a side note, the 0-10 slice would lead to errors starting from the second day, but so far I just used the first day for verification._

Peter9192:
Comparison to latest master branch confirms the changes are introduced in this branch.

preprocess_ecearth.py
```diff
@@ -53,373 +23,213 @@
...
-            latnrs,
-            lonnrs,
-        ]  # kg/kg #:267,134:578
-        time_q2m = get_input_data("Q2M", year, month).variables["time"][
-            begin_time * 2 + 1 : (begin_time * 2 + count_time * 2 + 1) + 1 : 2
-        ]
-
-        # surface pressure is 3-hourly data (03.00,06.00,09.00,12.00,15.00,18.00,21.00,00.00)
-        lnsp = get_input_data("LNSP", year, month).variables["LNSP"][
-            begin_time * 2 : (begin_time * 2 + count_time * 2 + 1) : 2,
```
Peter9192, line 148: Note: this indexing is different from u10, q2m etc:

* q2m: `begin_time * 2 + 1 : (begin_time * 2 + count_time * 2 + 1) + 1 : 2` --> 1:10: 2 = [1, 3, 5, 7, 9]
* lnsp: `begin_time * 2 : (begin_time * 2 + count_time * 2 + 1) : 2` --> 0:9:2 = [0, 2, 4, 6, 8]

where I added how this evaluates for the first day.

preprocess_ecearth.py
```diff
@@ -102,46 +123,39 @@ def getWandFluxes(
...
+    v = join_levels(v, v10)
+    q = join_levels(q, q2m)
 
-    # Make pressure array
-    p = np.zeros((time, levelist.size + 2, len(latitude), len(longitude)))
-    p[:, 1:-1, :, :] = levelist[np.newaxis, :, np.newaxis, np.newaxis]
-    p[:, 0, :, :] = sp
+    u = repeat_top_level(u)  # top level will be set to 10000 hPa
+    v = repeat_top_level(v)
+    q = repeat_top_level(q)
```
Peter9192, line 115: Note: in the original code zeros are inserted for q, instead of repeating the top level.

Peter9192:
Okay, all code changes identified. In conclusion, there were the following changes:

- I accidentally used 6-hourly indexing for u10, v10.
- Original `lnsp` indexing was off. I had overlooked this and "accidentally" corrected it.
- I had repeated the top level `q` values whereas the original code padded zeros.

All behaviour is now back to original (so **with** potential bugs!) to maintain consistent behaviour and allow verification for subsequent changes. I'll open a new issue pointing out these issues.

Peter9192 (commented):
Looked at this together with Imme. Also ran the verification script to ensure that the output is still the same as the original code. Further work will be done in follow-up PRs.

There's a couple of fixes that still need to be applied here, but that will change the behaviour:

- reading of multiple timesteps, this needs to be reorganized. Currently, it reads only the first 5 timestamps, so for later times the output is wrong.
- fixing lsnp timesteps
- inserting the pressure boundary check bottom layer zeros

Followup PR will focus on syncing EC-Earth and ERA5 preprocessing

PR #40, opened 2022-05-20
Add new era5 flood case

Peter9192:
and further simplify preprocessing script


Peter9192:
I'll merge this as it will make working with it next week easier.

PR #43, opened 2022-06-20
Added preprocess file ERA5 handles model levels

Imme1992:
Added script that reads in model level data ERA5

### DIFF
--- a/preprocess_era5_modellevels.py
+++ b/preprocess_era5_modellevels.py
@@ -14,7 +14,8 @@ g = 9.80665  # [m/s2]
 density_water = 1000  # [kg/m3]
 
 # Select model levels
-modellevels = [1,20,40,60,80,100,110,120,125,130,131,132,133,134,135,136,137]
+modellevels = np.append( np.arange(1,100,10), np.arange(101,138,1)) #[1,20,40,60,80,100,110,120,125,130,131,132,133,134,135,136,137]
+print('Number of model levels:',len(modellevels))
 
 # Read case configuration
 with open("cases/era5_2013.yaml") as f:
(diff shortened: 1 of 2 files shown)


preprocess_era5_modellevels.py
```diff
@@ -0,0 +1,178 @@
...
+    
+    print(
+        "Check calculation water vapor with column water vapour, this value should be close to zero:",
+        (cwv.sum(dim="lev") - (tcwm3)).sum().values,
+    )
+    # is niet hetzelfde.. hoe erg is dat?
+    
+    # Change units to m3
+    # TODO: Check this! Change units before interp is tricky, if not wrong
+    total_seconds = config["timestep"] / config["divt"]
```
chrisweij, line 121: timestep and diet have been deleted from the config file, there is now 1 config variable called tartet_freq (see preprocessing script era5 for pressure levels): 

Suggested code:
    # Change units to m3, based on target frequency (not incoming frequency!)
    target_freq = config['target_frequency']
    total_seconds = pd.Timedelta(target_freq).total_seconds()

preprocess_era5_modellevels.py
```diff
@@ -0,0 +1,178 @@
...
+        w_upper,
+        w_lower,
+    )
+
+    # Save preprocessed data
+    # Note: fluxes (dim: time) are at the edges of the timesteps,
+    # while states (dim: time2) are at the midpoints and include next midnight
+    # so the first state from day 2 will overlap with the last flux from day 1
+    filename = f"{date.strftime('%Y-%m-%d')}_fluxes_storages.nc"
+    output_path = os.path.join(config["interdata_folder"], filename)
```
chrisweij, line 165: interdata_folder has been renamed to preprocessed_data_folder

output_path = os.path.join(config["preprocessed_data_folder"], filename)

preprocess_era5_modellevels.py
```diff
@@ -0,0 +1,178 @@
...
+    filename = f"FloodCase_201305_ml_{variable}.nc"
+    filepath = os.path.join(config["input_folder"], filename)
+    da = xr.open_dataset(filepath)[variable]
+
+    # Include midnight of the next day (if available)
+    extra = date + pd.Timedelta(days=1)
+    return da.sel(time=slice(date, extra)).sel(lev=modellevels)
+
+
+datelist = pd.date_range(
```
chrisweij, line 45: "start_date" and "end_date" have been renamed to "preprocess_start_date" and "preprocess_end_date" in the config file.

Suggested change
datelist = pd.date_range(
    start=config["preprocess_start_date"], end=config["preprocess_end_date"], freq="d", inclusive="left"
)

chrisweij (approved):
Code works with suggested changes, there are a few things that should be resolved as well (add units in comments to be consistent with pressure levels preprocessing file, and the automatically calculation of dp for different model levels), but those work better in a separate issue.

PR #51, opened 2022-06-22
restructured settings file

### LINKED ISSUE #52: Restructure settings file
- Divide in settings for preprocessing & tracking
- Remove region and other unused settings
- ...

ruudvdent:
closes #52

### DIFF
--- a/cases/era5_2013.yaml
+++ b/cases/era5_2013.yaml
@@ -1,7 +1,7 @@
 # Settings needed for the data preprocessing (used by preprocess_XXXXXX.py)
 input_folder: ../era5_2013
-preproces_start_date: '20130521' #YYYYMMDD
-preproces_end_date: '20130604' #YYYYMMDD
+preprocess_start_date: '20130521' #YYYYMMDD
+preprocess_end_date: '20130604' #YYYYMMDD
 
 #######################
 # divt & count_time # TODO: Peter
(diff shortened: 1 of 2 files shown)


cases/era5_2013.yaml
```diff
@@ -1,19 +1,30 @@
...
+
+# Settings needed to define the tracking region (in space and time)
+region: ../era5_2013/FloodCase_201305_lsm.nc # requirements: a) same extent in space as preprocessed data, c) values between 0 and 1
+track_start_date: '20130521' #YYYYMMDD
+track_end_date: '20130604' #YYYYMMDD
+
+# Settings needed for the tracking run
+name_of_run: 'default_run'
+veryfirstrun: true # true: starts from zero tracked water, false: loads tracked water
+Kvf: 3 # Vertical transport parameter for gross vertical transport between the layers during the tracking: "actual exchange = Kvf * F_vertical + F_vertical" in one direction and "-1 * (Kvf * F_vertical)" in opposite direction. # Default = 3.
```
Peter9192, line 23: This comment is quite elaborate. Someday/maybe it might fit better in the documentation, with a short note/pointer in the inline code

Peter9192 (approved):
Looks good to me! I verified that the code works, but only if the `preprocessed_data` path exists. 

Shall we update the data paths to those on the SURF machine?

cases/era5_2013.yaml
ruudvdent: could indeed be part of readme, but not priority now

PR #57, opened 2022-06-22
Create preproc data if not exists

### LINKED ISSUE #53: create pre_processed dir if not exists

Peter9192:
Is that what you had in mind @ruudvdent?

closes #53


ruudvdent:
gives error if directory already exists, can you check Peter

Peter9192:
> gives error if directory already exists, can you check Peter

Should be fixed now. Can you verify?

ruudvdent (approved):
indeed solved

PR #78, opened 2022-06-27
linked the stability factor vertical to kvf in config

ruudvdent:
linked the stability factor vertical to kvf in config


Peter9192 (approved):
Thanks, that's a good idea!

PR #80, opened 2022-07-02
Backtrack

Peter9192:
This PR further refactors the code for backtracking. The whole sample case now runs on my laptop in under 10 seconds. Some of the important changes are:

* Abbreviate commonly used code (e.g. `shift_east(...)`, `[inner]` and `s_track_relative`)
* Remove as many as possible empty array allocations
* Do the daily accumulations "on the fly", i.e. don't store intermediate values (3d) and then sum them, but just add them directly to a single (2d) output array.
* Compute intermediate variables (e.g. fx_e_lower_we) only one time step at a time. That saves a lot of working memory.

Please review thoroughly to identify any mistakes I might have made.

### DIFF
--- a/backtrack.py
+++ b/backtrack.py
@@ -6,7 +6,7 @@ from pathlib import Path
 import pandas as pd
 
 # Read case configuration
-with open("cases/era5_2013.yaml") as f:
+with open("cases/era5_2013_local.yaml") as f:
     config = yaml.safe_load(f)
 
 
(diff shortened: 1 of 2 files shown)


backtrack.py
```diff
@@ -31,424 +31,243 @@ def input_path(date):
...
-    states = xr.Dataset(
+    # Pack processed data into new dataset
+    ds = xr.Dataset(
         {
-            "s_track_upper": (["time_states", "lat", "lon"], s_track_upper),
-            "s_track_lower": (["time_states", "lat", "lon"], s_track_lower),
+            "s_track_upper_restart": (["lat", "lon"], s_track_upper),  # Keep last state for a restart
+            "s_track_lower_restart": (["lat", "lon"], s_track_lower),
+            "s_track_upper": (["lat", "lon"], s_track_upper_mean),
+            "s_track_lower": (["lat", "lon"], s_track_upper_mean),
```
Imme1992, line 229:
```suggestion
            "s_track_lower": (["lat", "lon"], s_track_lower_mean),
```

Imme1992 (commented):
Some small comments on naming variables etc. and one bug where s_track_lower was named s_track_upper

Imme1992:
I think there are still negative values in e_track left as the evaporation variable itself can still be negative, preprocess script is not the same script as in Master branch (where evaporation is made always positive)

Imme1992 (approved):

ruudvdent:
todo: re-check negative values after updated pre-processing from backtrack

PR #82, opened 2022-07-19
Corrected column and fluxes to include cloud water + ice

Imme1992:
Hello, I adapted the model level code so that the integrated water column is corrected for not consisting of cloud water and cloud ice. However, I am not sure how to correct the fluxes for only consisting of water vapour and not including cloud water and cloud ice, so the code for that is still missing/not correct.


Imme1992:
@ruudvdent I am looking into changing the code to adjust for water vapour vs water. I want now to change from column water vapour to column water and then calculate the fluxes based on the column water. 
What I do not understand yet that in the original code the vapor_top and vapor_down are first calculated from the column water vapour and then adjusted for to water (W_top and W_down line 163 +164 in Fluxes_and_States_Masterscript.py) , but in principle you could also directly calculate water_top and water_down from integrating the cwv variable right?

preprocess_era5_modellevels.py
```diff
@@ -86,48 +83,78 @@ def load_modellevel_data(variable, date):
...
+    
+        # Determine the states
+        cwv = q * dp_modellevels / g  #  # column water vapor (kg/m2)
    
-    print(
-        "Check calculation water vapor over two layers, this value should be zero:",
-        (cwv.sum(dim="lev") - (w_upper + w_lower)).sum().values,
-    )
+        # calculate column water instead of column water vapour
+        cw = (tcw / cwv.sum(dim="lev")) * cwv #column water (kg/m2)
```
chrisweij, line 122: Looks okay, but this changes the shape of the resulting array:
cw has the shape [time,lat,lon,lev], while the original array cwv has the shape [time,lev,lat,lon]. 

Might not be problematic but good to check if this causes any problems.

preprocess_era5_modellevels.py
```diff
@@ -86,48 +83,78 @@ def load_modellevel_data(variable, date):
...
-    w_lower = cwv.where(lower_layer).sum(dim="lev")
-
-    fa_e_upper = fa_e.where(upper_layer).sum(dim="lev")
-    fa_n_upper = fa_n.where(upper_layer).sum(dim="lev")
-    w_upper = cwv.where(upper_layer).sum(dim="lev")
+        
+    if config['vertical_integral_available'] == True:
+        #load total water column
+        tcw = load_surface_data("tcw", date) # kg/m2
+        #vieclw = load_surface_data("vieclw", date) # kg m-1 s-1
```
chrisweij, line 109: Are these variables still necessary?

chrisweij (approved):
Looks fine, one thing that should be changed as well is adding the 'periodic_boundary' and 'kvf' variables to the call of the function fa_vert. I could not add this as a line comment, but it is line nr. 187 and further. Now it gives an error (probably changed in another pull request), see code below. 

    # Determine the vertical moisture flux
    fa_vert = get_vertical_transport(
        fa_e_upper,
        fa_e_lower,
        fa_n_upper,
        fa_n_lower,
        evap,
        precip,
        w_upper,
        w_lower,
        config['periodic_boundary'],
        config['kvf']
    )

preprocess_era5_modellevels.py
Imme1992: No not in the way we do it now, I deleted them and will commit after

Imme1992:
> Looks fine, one thing that should be changed as well is adding the 'periodic_boundary' and 'kvf' variables to the call of the function fa_vert. I could not add this as a line comment, but it is line nr. 187 and further. Now it gives an error (probably changed in another pull request), see code below.

Yes indeed, I added those lines

preprocess_era5_modellevels.py
Imme1992: Yes I see, it causes no problems

PR #90, opened 2022-07-27
Floodcase july2021

Imme1992:
I am working on the case of July 2021, tracking the extreme precip
I have corrected a few things:
    north_loss = np.zeros(nlon)
    south_loss = np.zeros(nlon)
    east_loss = np.zeros(nlat) 
    west_loss = np.zeros(nlat)

I have added the option to track for certain events/days instead of the full climatology


Imme1992:
I do not want to merge this branch in master necessarily but just publish the branch so it is available for others
In this tracking script you can select the event that you want to track

Peter9192 (approved):
Well done @Imme1992! I'm happy that you found some of my errors, and that we managed to have the first meaningful results :-)

I polished it a bit more, moving the plotting functions to a separate script. Everything now works on my side as well.

Will continue to restructure the code in follow-up PRs.

PR #99, opened 2022-08-12
Dp calc

ruudvdent:
Just created to diff the file, not an actual pull request


ruudvdent:
we do not want to pull these changes (done in another branch)

PR #100, opened 2022-08-12
Calculate q2m and q,u,v, interpolation between top-boundary-surface

ruudvdent:
surface calculation by Chris (was already reviewed) combined with a more precise q,u,v, interpolation between top-boundary-surface (but the latter is slow)


Peter9192 (commented):
I see the struggle! So if I understand correctly, this code adds three additional levels:

- Top of atmosphere (insert dummy value)
- Boundary pressure level (interpolate between surrounding values at each grid point)
- Surface (insert in between the "surrounding" pressure levels at each point)

Data at pressure levels below the surface will be 0 (?) Or shall we make it nan?

Is that correct? I think this might be possible without the loop :-). Let me look into it

ruudvdent:
_The three additional levels are:

    Top of atmosphere (insert dummy value)_ - not a a dummy value, it is assumed q = 0 and u and v are equal to the closest pressure levels
    _Boundary pressure level (interpolate between surrounding values at each grid point)_ - correct
    _Surface (insert in between the "surrounding" pressure levels at each point)_ - correct, and any additional array elements that are 'below' the surface also get surface pressure so that the dp between these elements is zero.

Dummy values of 1100 hPa were used temporarily in to calculate the location of the surface and boundary

_Data at pressure levels below the surface will be 0 (?) Or shall we make it nan?_  - in this code the data for q, u, v below surface gets assigned the surface value, however, this does not really matter as dp will be calculated to be zero.

Is that correct? I think this might be possible without the loop :-). Let me look into it - excellent

Peter9192:
Please see #103 and use it if you like it.

Pro tip: you can quote text from issues by selecting the lines you want, click the three dots in the right top, and press "quote reply". Alternatively, you can use the `>` sign to make a line look like a quote:

> this looks like a quote.

Peter9192 (approved):
I'm going to merge this so we can continue. Please check later.

PR #101, opened 2022-08-15
added keywords

ruudvdent:


Peter9192:
I validated the citation file with https://pypi.org/project/cffconvert/, and found that there's an error. `orc-id` should be changed to `orcid`. Can you add that to this PR?

Peter9192 (approved):
Fixed and validated again.

PR #102, opened 2022-08-15
added version and release data to citation.cff

ruudvdent:


ruudvdent:
The doi would also be nice-to-have (I need to cite this repository in the coming weeks)

Peter9192:
I we want a DOI, we need to make a release. We discussed earlier (#4) that we'd make a first v3 release once we had settled a bit more on a stable version, but we're still making quite a few changes. Shall we discuss this on Friday?

Peter9192 (approved):
doi with "todo" field not valid cff. Also had to change the date format. Now all is ok. We can add the DOI after the first official release.

PR #108, opened 2022-08-19
Synchronize preproc for pressure and model levels

Peter9192:
Started an attempt to sync preprocessing for model- and pressure levels. Work in progress.

In the process, I moved all unit changes to the tracking (insofar that was not done yet).

Notes:

- pressure levels are descending, model levels ascending?
- level dimension is called "lev" in one file and "level" in the other
- copied total column water correction so it is the same in both files

PR #109, opened 2022-08-19
debugged pressure level calc

ruudvdent:


Peter9192:
I copied all these edits over to #108, so closing this one.

PR #113, opened 2022-08-22
Add functions to check pre-processed data

Peter9192:
This adds several utility functions to check all kinds of things related to (pre-processed) data. Several aspects are up for discussion:

* Should we issue warnings or raise exceptions?
* How to deal with different notations of units?
* Do we agree on the current characteristics of input data, as summarized at the bottom of `preprocessing/checks.py`?
* What to do with the extra midnight timestamp?

To do:

* [x] Also check output of pressure level script
* [x] Point to data requirements in documentation
* [x] Check regularity of grid spacing and time?
* [x] Change precip and evap to fluxes
* [x] Check latitude only decreases

Updated documentation: https://wam2layers--113.org.readthedocs.build/en/113/preprocessing.html

### DIFF
--- a/wam2layers/preprocessing/checks.py
+++ b/wam2layers/preprocessing/checks.py
@@ -114,12 +114,12 @@ VARIABLES = yaml.safe_load("""
         all_positive: True
         coordinates: [time, latitude, longitude]
     s_upper:
-        valid_units: ["kg m-2", "kg/m2", "kg m-2 s-1", "kg/m2/s"]
+        valid_units: ["kg m-2", "kg/m2"]
         range: [0, 1000]
         all_positive: True
         coordinates: [time, latitude, longitude]
     s_lower:
-        valid_units: ["kg m-2", "kg/m2", "kg m-2 s-1", "kg/m2/s"]
+        valid_units: ["kg m-2", "kg/m2"]
         range: [0, 1000]
         all_positive: True
         coordinates: [time, latitude, longitude]
--- a/wam2layers/preprocessing/checks.py
+++ b/wam2layers/preprocessing/checks.py
@@ -1,10 +1,11 @@
 import warnings
 from pathlib import Path
 
+import numpy as np
 import yaml
 
 
-def warning_on_one_line(message, category, filename, lineno, file=None, line=None):
+def _warning_on_one_line(message, category, filename, lineno, file=None, line=None):
     """A more compact warning format.
 
     https://stackoverflow.com/a/26433913
@@ -14,7 +15,7 @@ def warning_on_one_line(message, category, filename, lineno, file=None, line=Non
 
 
 # Override the format with which warnings are printed
-warnings.formatwarning = warning_on_one_line
+warnings.formatwarning = _warning_on_one_line
 
 
 def check_monotonic_increase(data, dim="level"):
@@ -77,11 +78,19 @@ def check_range(data, range):
         warnings.warn(f"Data is not within the range {min} to {max}.")
 
 
+def check_uniform(coord):
+    spacing = np.diff(coord)
+    if not spacing.min() == spacing.max():
+        warnings.warn(f"Coordinate spacing is not uniform for coord {coord}")
+
+
 def check_coords(data, coords):
     check_shape(data, ndim=len(coords))
     for coord in coords:
         if coord not in data.coords:
             warnings.warn(f"Data does not have a coordinate {coord}.")
+        else:
+            check_uniform(data[coord])
 
 
 def check_input(data):
--- a/wam2layers/preprocessing/checks.py
+++ b/wam2layers/preprocessing/checks.py
@@ -1,0 +1,137 @@
+import warnings
+from pathlib import Path
+
+import yaml
+
+
+def warning_on_one_line(message, category, filename, lineno, file=None, line=None):
+    """A more compact warning format.
+
+    https://stackoverflow.com/a/26433913
+    """
+    short_path = Path(filename).relative_to(Path(filename).parents[2])
+    return f'{short_path}:{lineno}: {category.__name__}: {message}\n'
+
+
+# Override the format with which warnings are printed
+warnings.formatwarning = warning_on_one_line
+
+
+def check_monotonic_increase(data, dim="level"):
+    """Check if data is monotonically increasing in the given dimension."""
+    if not data.diff(dim).min() >= 0:
+        warnings.warn(f"Data is not monotonically increasing in the dimension {dim}.")
+
+
+def check_monotonic_decrease(data, dim="level"):
+    """Check if data is monotonically decreasing in the given dimension."""
+    if not data.diff(dim).max() <= 0:
+        warnings.warn(f"Data is not monotonically decreasing in the dimension {dim}.")
+
+
+def check_units(data, valid_units):
+    """Check if data has the given units."""
+    if "units" not in data.attrs:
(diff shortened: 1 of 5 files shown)


wam2layers/preprocessing/checks.py
```diff
@@ -0,0 +1,146 @@
...
+
+
+def _warning_on_one_line(message, category, filename, lineno, file=None, line=None):
+    """A more compact warning format.
+
+    https://stackoverflow.com/a/26433913
+    """
+    short_path = Path(filename).relative_to(Path(filename).parents[2])
+    return f'{short_path}:{lineno}: {category.__name__}: {message}\n'
+
```
Imme1992, line 15: I do not really get what this warning_on_one_line does? What is the use of this function

wam2layers/preprocessing/checks.py
```diff
@@ -0,0 +1,146 @@
...
+# Override the format with which warnings are printed
+warnings.formatwarning = _warning_on_one_line
+
+
+def check_monotonic_increase(data, dim="level"):
+    """Check if data is monotonically increasing in the given dimension."""
+    if not data.diff(dim).min() >= 0:
+        warnings.warn(f"Data is not monotonically increasing in the dimension {dim}.")
+
```
Imme1992, line 25: Is this to check if the latitudes are increasing or decreasing?

wam2layers/preprocessing/checks.py
```diff
@@ -0,0 +1,146 @@
...
+def check_input(data):
+    for variable, var_info in VARIABLES.items():
+        check_units(data[variable], var_info["valid_units"])
+        check_coords(data[variable], var_info["coordinates"])
+        check_range(data[variable], var_info["range"])
+        if var_info["all_positive"]:
+            check_positive(data[variable])
+
+
```
Imme1992, line 104: This checking of the variables range and units is a good check, nice

wam2layers/preprocessing/preprocess_era5_modellevels.py
```diff
@@ -98,29 +101,28 @@ def load_data(variable, date, levels=None):
...
-    s_lower = cw.where(lower_layer).sum(dim="lev")
-    s_upper = cw.where(upper_layer).sum(dim="lev")
+    s_lower = cw.where(lower_layer).sum(dim="lev", keep_attrs=True)
+    s_upper = cw.where(upper_layer).sum(dim="lev", keep_attrs=True)
 
     # Determine the fluxes
-    fx = u * cw  # eastward atmospheric moisture flux (kg m-1 s-1)
-    fy = v * cw  # northward atmospheric moisture flux (kg m-1 s-1)
+    fx = (u * cw).assign_attrs(units="kg m-1 s-1")  # eastward atmospheric moisture flux (kg m-1 s-1)
+    fy = (v * cw).assign_attrs(units="kg m-1 s-1")  # northward atmospheric moisture flux (kg m-1 s-1)
```
Imme1992, line 117: Is it not a bit tricky to assign units? In case the units are not correct? But when you do the check before it should be fine

wam2layers/preprocessing/preprocess_era5_modellevels.py
```diff
@@ -131,4 +133,12 @@ def load_data(variable, date, levels=None):
             "evap": evap,
             "precip": precip,
         }
-    ).to_netcdf(output_path)
+    )
+
+    # Verify that the data meets all the requirements for the model
+    check_input(ds)
+
```
Imme1992, line 140: You check the input here while you assigned units before, is that not a strange order? I mean ofcourse when you assign the units the unit check will be fine. Or do I misunderstand?

Imme1992 (approved):

Imme1992:
I think this changes are especially useful when a different dataset then ERA5 is used. If we 'just' have the ERA5 data to me the checks are not needed, but when using another dataset this checks would be good to do.

wam2layers/preprocessing/checks.py
Peter9192: This a bit technical indeed, I'd just skip over it if I were you. What it does: make the formatting of the warning messages a bit nicer. Now it prints something like: `wam2layers/preprocessing/checks.py:35: UserWarning: precip has no units`

wam2layers/preprocessing/checks.py
Peter9192: Not whether it is in- or decreasing, but rather that it is only, and consistently, increasing. So

* [0, 1, 2, 4, 7] would pass
* [3, 2, 1] would fail
* [1, 2, 4, 3] would fail as well

wam2layers/preprocessing/preprocess_era5_modellevels.py
Peter9192: It is. Unfortunately, xarray doesn't check units when doing maths. It happily let's you add metres to kilograms. There is a similar package, iris, that is much stricter on this. However, in my experience, it is quite a bit more difficult to work with.

By adding this check and adding units in general, I hope that users will be reminded of the fact that they need to be careful. However, it remains their own responsibility.

Peter9192: When you do the check before, it fails, because there are no units at all...

wam2layers/preprocessing/preprocess_era5_modellevels.py
Peter9192: You need to assign the units somewhere... xarray often discards them or they get outdated when we do math operations. I want to do the check at the very end, so it acts as a final verification that our preprocessed data meets the requirements for tracking.

Peter9192:
Thanks for reviewing!

wam2layers/preprocessing/checks.py
```diff
@@ -0,0 +1,146 @@
...
+    s_upper:
+        valid_units: ["kg m-2", "kg/m2", "kg m-2 s-1", "kg/m2/s"]
+        range: [0, 1000]
+        all_positive: True
+        coordinates: [time, latitude, longitude]
+    s_lower:
+        valid_units: ["kg m-2", "kg/m2", "kg m-2 s-1", "kg/m2/s"]
+        range: [0, 1000]
+        all_positive: True
+        coordinates: [time, latitude, longitude]
```
ruudvdent, line 125: not per second

wam2layers/preprocessing/checks.py
```diff
@@ -0,0 +1,146 @@
...
+    evap:
+        valid_units: ["m", "m of water equivalent"]
+        range: [0, 0.001]
+        all_positive: True
+        coordinates: [time, latitude, longitude]
+    precip:
+        valid_units: ["m", "m of water equivalent"]
+        range: [0, 0.1]
+        all_positive: True
+        coordinates: [time, latitude, longitude]
```
ruudvdent, line 115: better as fluxes, otherwise it depends on time step which makes the check quite difficult.

Peter9192: Done in https://github.com/WAM2layers/WAM2layers/pull/113/commits/6afa129e07419eb6371d1cfb2a4190de0612372e

wam2layers/preprocessing/checks.py
Peter9192: Fixed in https://github.com/WAM2layers/WAM2layers/pull/113/commits/7a3bde20ca262d48bc5d6586e39b3b95e5e2f03e

PR #114, opened 2022-08-24
Update CITATION.cff

ruudvdent:
added Chris and fixed family name issue ("van der" to family name instead of name-particle, which messes up things when citing)


Peter9192 (approved):
Thanks!

Issue #110, opened 2022-08-19
Run global case to see if the code is still suitable (thus also download ERA5 globally)


ruudvdent:

Imme1992:
Done, this works now see also issue #136

Issue #91, opened 2022-08-05
Correct calculation of dp if pressure levels


chrisweij:
The calculation of dp using pressure levels instead of model levels should be checked.

- First, it should be checked/changed that the total dp over all levels is equal to the total column height (from surface pressure to 0 hPa?)
- Secondly, the calculation of dp should be accounted for when separating the data into two layers

ruudvdent:
I think this can in principle be closed per #109, which does this correctly, although the code does not provide feedback on whether it's correct. Do you want to add that @chrisweij ?

Imme1992:
@Peter9192 has local adjustments which still need to be included

Imme1992:
Can you also document this in a log file? What is the correction applied when tcw is used

ruudvdent:
![image](https://github.com/WAM2layers/WAM2layers/assets/10234007/fca98b7b-848c-4ad5-98a5-cb21ae225a15)

ruudvdent:
![image](https://github.com/WAM2layers/WAM2layers/assets/10234007/6bca8030-6fc0-4e08-9aff-cb9ac3156e6a)

Issue #46, opened 2022-06-20
Allow for global simulations in tracking


Imme1992:
In Backward and Forward Tracking add the global simulations option (see settings script and pre-processing for similar code)

Peter9192:
e.g. by adding a setting: periodic=True

ruudvdent:
fixed for preprocessing, but reopened for backward tracking script

Imme1992:
This still need to be checked carefully

Imme1992:
Same as issue #136 
Branch created Global_runs where periodic boundaries are coded in the backtracking script

Imme1992:
Dask needs to be properly installed and used within the program
Dask need to be included in the package installation
Finetuning of loading of global data using dask (make dask optional)

Peter9192:
Possible though might need dask options added in #119 for big arrays.

PR #119, opened 2022-09-02
Allowing global input data

### LINKED ISSUE #117: Standardize input filename pattern
If we want to be more flexible to different types of input data, it would be helpful to have a standardized data loading function. This function needs to know about the organization of your data on your disk. One option would be to have this fixed. Alternatively, we can make it flexible by passing in a filename pattern. We suggest the following default filename pattern:

```
filename_pattern: "{ERA5}/{Year}/{Month}/{ERA5}_{Year}_{Month}_{Day}_{Variable}{levttype}.nc"  # levtype = "_ml" or "_pl" or ""
```

### LINKED ISSUE #126: Use daily input data for preprocessing
So far we have run cases with one variable per input file, but all dates in one file. For bigger experiments it would be more flexible if we save the data in daily files. For this, we need to adapt the code.

Related to #117

Imme1992:
We have changed the preprocess script so it allows to load daily files.
Dat is now loaded in chunks so there is no memory issue when global data is loaded, however this makes the code slower (thus it would be good to find a solution to decide to chunk or not) 
First, we included the midnight timestep from the next day, but the current script does not do this. If we want to continue wit h this script we have to change the backtracking code.

To do:
- [ ] Make sure that chunking is only done when necessary

closes #126 
closes #117


Peter9192:
In #131 we adapted the code such that it should be able to run without the additional midnight.

Peter9192:
This branch has become a bit messy. Let's split it up into several small PRs.

- [ ] Periodic boundaries #136
- [ ] Reading global data in chunks so preproc works on a machine with small memory #136 #46 #110
- [ ] Filename template for raw input data & no need to include next midnight #117 (optionally also include #137)
- [ ] Fix for era5 modellevels dp calculation #91

Peter9192:
All changes from this PR have been merged through other PRs, except for the dask chunking. I've made it such that you can set a `chunks` argument in the configuration file. 

```yaml
# don't use dask:
chunks: null   

 # process one time step at a time
chunks: {time: 1} 

# set chunking for all individual dims
chunks:  
    level: -1
    time: 1  # don't set time to auto, as it will be different for surface and 3d vars, that's bad for performance
    latitude: auto
    longitude: auto
```

@Imme1992 can you test the preprocessing with the latest version of this branch to see if it works for you? You will need to make sure to set the `chunks` argument in your configuration file. I recommend the second option (` chunks: {time: 1} `)

Please also update your environment beforehand, by running `pip install --update --editable .` to make sure you have the latest versions of dask and distributed.

Imme1992:
Hi @Peter9192 , I checked it and get the error
KeyError: 'use_dask'

Do I need to add in the config:
use_dask: True
?

Imme1992:
@Peter9192 when I have added the use_dask: True in the configuration script it works, I do however get quite some errors/warnings that I do not understand, see here for a screenshot
image

Imme1992:
@Peter9192 I also get an error at the last timestep when doing the preprocessing using chunks, for a screenshot see below:
image

Imme1992:
I just checked the output from the chunked preprocessed and not chunked preprocessing and it looks very different for timestep 2021-07-14.

Peter9192:
> Hi @Peter9192 , I checked it and get the error
> KeyError: 'use_dask'

That's unexpected... Can you copy the full (or bigger part of) the error message? Which line triggers this?

Peter9192:
> I just checked the output from the chunked preprocessed and not chunked preprocessing and it looks very different for timestep 2021-07-14. (the values of fx_upper in the chunked version are averaged over time, latitude and longitude for the 14th of July negative, and positive for the original preprocess outputs)

That's not good. Let me get back to this

Imme1992:
Regarding my previous comments here I made the mistake to not pull the latest version. 
Now I have rerun with the latest version of the branch, and I don't get intermediate errors, and the runtime is much faster. Only in the last time step I get the follwing error:
image

I checked the output data and it is similar as the output data without the dask chunking.
The checking of units is not done in this pull request as the code needs to be rewritten to handle chunked (dasked) arrays.

Peter9192 (approved):
Let's address remaining issues (code works but throws lots of warnings/errors) in follow-up PRs

PR #120, opened 2022-09-02
Download script

chrisweij:
Automatic download data script

Still to do
- Directory structure should be added as planned. 
- Additional single levels at the surface if downloading pressure levels (u10, v10, t2m,..)
- Function to standarize downloading, independent of level type?


chrisweij:
Directory structure is as planned, and extra data is downloaded as well for pressure levels. Standarized function to download data is probably not that convenient (c.retrieve is already a function). 

Only thing is that it now downloads daily data files, might be better to download the data per month (and split the files afterwards). But for that you should a priori know if the files are not too big.

cases/era5_2021.yaml
```diff
@@ -1,5 +1,6 @@
 # Settings needed for preprocessing
-input_folder: /data/volume_2/era5_2021
+#input_folder: /data/volume_2/era5_2021
+input_folder: /data/volume_2/ERA5
```
Peter9192, line 3:
```suggestion
input_folder: /data/volume_2/ERA5
```
I'd suggest to just remove the outdated line. If we ever need it back, it is still available in the git history

cases/era5_2021.yaml
```diff
@@ -25,3 +26,8 @@ distancetracking: false
 
 event_start_date: '20210713'
 event_end_date: '20210715'
+
+# Settings for downloading the data
+data_freq: 'hourly'
+data_area: [60, -50,30,30] # North, West, South, East. Default: global
```
Peter9192, line 32:
```suggestion
data_area: [60, -50, 30, 30]  # North, West, South, East. Default: global
```
Perhaps you can document these settings also in docs/config.md ?

get_era5.py
```diff
@@ -0,0 +1,157 @@
+import cdsapi
```
Peter9192, line 1: Shall we place `get_era5.py` in `src/wam2layers/download/era5.py` or `src/wam2layers/utils/download_era5.py`?

get_era5.py
```diff
@@ -0,0 +1,157 @@
...
+c = cdsapi.Client()
+
+# Read case configuration
+with open("cases/era5_2021.yaml") as f:
+    config = yaml.safe_load(f)
+    
+# Use the preprocessing dates to select which dates to download
+datelist = pd.date_range(
+    start=config["preprocess_start_date"], end=config["preprocess_end_date"], freq="d"
+)
```
Peter9192, line 15: perhaps we can re-use some code from the preprocessing module:

```suggestion
from wam2layers.preprocessing.era5 import parse_config
config=parse_config("cases/era5_2021.yaml")
```

Eventually I think it would be nice to have "parse_config" in a separate module, like `wam2layers.utils`.

get_era5.py
```diff
@@ -0,0 +1,157 @@
...
+    s = '/'
+    levlist = s.join(str(e) for e in config['levels'])
+
+# Construct list of downloading times
+if (config['data_freq'] == 'hourly'):
+    timlist  = '00:00:00/01:00:00/02:00:00/03:00:00/04:00:00/05:00:00/06:00:00/07:00:00/08:00:00/09:00:00/10:00:00/11:00:00/12:00:00/13:00:00/14:00:00/15:00:00/16:00:00/17:00:00/18:00:00/19:00:00/20:00:00/21:00:00/22:00:00/23:00:00'
+elif(config['data_freq'] == '3hourly'): 
+    timlist = '00:00:00/03:00:00/06:00:00/09:00:00/12:00:00/15:00:00/18:00:00/21:00:00'
+elif(config['data_freq'] == '6hourly'): 
+    timlist = '00:00:00/06:00:00/12:00:00/18:00:00'
```
Peter9192, line 32: do we ever want to use/support 3- or 6-hourly data?

get_era5.py
```diff
@@ -0,0 +1,157 @@
...
+        dtstr = date.strftime("%Y-%m-%d")
+
+        #For first day of month, datadir should be created
+        data_dir = Path(config["input_folder"] + "/" + str(date.year) ).expanduser()
+        data_dir.mkdir(exist_ok=True, parents=True)
+        
+        #Subdirectory for every month 
+        data_dir = Path(config["input_folder"] + "/" + str(date.year) + '/' + str(date.month) ).expanduser()
+        data_dir.mkdir(exist_ok=True, parents=True)
+
```
Peter9192, line 59: Pathlib is very cool. Instead of making a path string with slashes (which might not work on windows), you can do:

`Path(input_folder).expanduser() / str(date.year) ` (note that the `/` is a division operator slash, not a string. 

So, you could reduce this to:

```suggestion
        # Create yearly parent dir if it doesn't exist yet
        parent = Path(config["input_folder"]).expand_user() / str(date.year)
        parent.mkdir(exist_ok=True, parents=True)
        
        # Create monthly subdirectories
        child = parent / str(date.month)
        child.mkdir(exist_ok=True)    # note: expanduser not necessary, already done above
```

get_era5.py
```diff
@@ -0,0 +1,157 @@
...
+        data_dir = Path(config["input_folder"] + "/" + str(date.year) + '/' + str(date.month) ).expanduser()
+        data_dir.mkdir(exist_ok=True, parents=True)
+        
+        #Loop over model level variables
+        for i in range(len(var_names)):
+            outfile =  'ERA5_' + dtstr + "_" + var_names[i] + '_ml.nc'
+            outfile = data_dir / outfile
+            
+            c.retrieve('reanalysis-era5-complete', {
+            'class': 'ea',
```
Peter9192, line 66: So if it fails midway, and I run it again, it will start again from the beginning, right? It would be nice (at some point, later), to make it check whether it already exists, and then skip it.

Peter9192 (commented):
Hi Chris, nice work! I had a quick look at the code (didn't test it yet). I left a few suggestions, feel free to ignore. 

One question: do we also want column integrated water or smt like that?

I agree that making a function around `c.retrieve` feels a bit seems convoluted. However, perhaps we can make two or three separate functions like `download_surface_var`, `download_pl_var` and `download_ml_vars`. If you do that, the outermost if-conditions would be a bit more compact, something like:

```py
if config['level_type'] == 'model_levels':
    download_model_levels(dates, [u, v, q])
    download_surface_vars(dates, [sp, tp, e])
else:
    download_pressure_levels(dates, [u, v, q])
    download_surface_vars(dates, [u10, v10, ...])
```

If you're out of time, we can keep it as is, and I could also give it a try later on.

get_era5.py
chrisweij: Yes, I could do that as well (though it might be nice to have a basic working download script first).

get_era5.py
chrisweij: I have moved it to src/wam2layers/utils/download_era5.py

chrisweij:
> Hi Chris, nice work! I had a quick look at the code (didn't test it yet). I left a few suggestions, feel free to ignore.
> 
> One question: do we also want column integrated water or smt like that?
> 
> I agree that making a function around `c.retrieve` feels a bit seems convoluted. However, perhaps we can make two or three separate functions like `download_surface_var`, `download_pl_var` and `download_ml_vars`. If you do that, the outermost if-conditions would be a bit more compact, something like:
> 
> ```python
> if config['level_type'] == 'model_levels':
>     download_model_levels(dates, [u, v, q])
>     download_surface_vars(dates, [sp, tp, e])
> else:
>     download_pressure_levels(dates, [u, v, q])
>     download_surface_vars(dates, [u10, v10, ...])
> ```
> 
> If you're out of time, we can keep it as is, and I could also give it a try later on.
> 
> PS: I merged master into this PR to make sure it incorporates the latest changes in master.

Thanks Peter, I did implement most of your suggestions, I think it might be possible to write the suggested functions. I can have a go a the end of the week.

get_era5.py
Peter9192: Let's open a new issue for that, so we can keep note of it but not do it here.

Peter9192:
> One question: do we also want column integrated water or smt like that?

Just to check: did you see this question?

chrisweij:
> > One question: do we also want column integrated water or smt like that?
> 
> Just to check: did you see this question?

I forgot to comment on it, but 'total column water' is a good idea (it is optional in the preprocessing script, but might be good to download standardly). I added to the list. I also fixed a few typo's and made the directory creation consistent for pressure levels.

chrisweij:
NB: I will also add something to the documentation, this should also include some reference to the cdsapi documentation of ECMWF

src/wam2layers/utils/get_era5.py
```diff
@@ -0,0 +1,152 @@
...
+        parent.mkdir(exist_ok=True, parents=True)
+        
+        # Create monthly subdirectories
+        child = parent / str(date.month)
+        child.mkdir(exist_ok=True)    # note: expanduser not necessary, already done above
+        
+        #Loop over model level variables
+        for i in range(len(var_names)):
+            outfile =  'ERA5_' + dtstr + "_" + var_names[i] + '_ml.nc'
+            outfile = data_dir / outfile
```
Peter9192, line 58: I think `data_dir` is now called `child`, no? Sorry for messing up the consistency... 
By the way, did you know about [f-strings](https://peps.python.org/pep-0498/)? You could also do:

```
outfile = f"ERA5_{dtstr}_{var_names[i]}_ml.nc"
```

I usually find those to be more readable than string concatenations.

chrisweij: Thanks, I forgot to change the variable names here. I also folllowed the recommendation and now I use f-strings (I always forgot that is an option). 

I also checked and it is still working for me.

Peter9192 (approved):

Peter9192:
Looks good, thanks!

PR #122, opened 2022-09-04
Update package for PyPi publication

### LINKED ISSUE #96: Should pandas>=1.4.0 be in requirements?
In the preprocessing script the parameter `inclusive` is used to create the date_range. This parameter was only introduced in version 1.4.0 from pandas (see https://pandas.pydata.org/docs/reference/api/pandas.date_range.html), so it might be necessary to set this as a requirement (or the code should be changed).

Peter9192:
Added some package metadata before publishing on PyPi.
Also closes #96 (but only after publication).

See https://pypi.org/project/wam2layers/

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -7,7 +7,7 @@ name = "wam2layers"
 description = "Atmospheric moisture tracking model"
 version = "3.0.0-beta.2"
 readme = "README.md"
-license = "Apache 2.0 licence"
+license = { text = "Apache 2.0" }
 classifiers = [
     "License :: OSI Approved :: Apache Software License",
     "Programming Language :: Python",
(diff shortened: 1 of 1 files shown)


Peter9192:
@chrisweij could you briefly check? This data ended up at PyPi, if there's something wrong or missing we can change it before the next release: https://pypi.org/project/wam2layers/

chrisweij (approved):
Hi Peter looks fine for me, 
pip install wam2layers worked both on my laptop (I did had to select the right python/pip version) as well the local supercomputer. I did also download the sample data, backtracking worked as well, but in the visualization step I got an "NotImplementedError" error message.

One small thing: in the documentation the yaml file is called floodCase_202107.yaml while in the GoogleDrive it is called floodcase_202107.yaml (without capital C).

chrisweij:
One other smile thing, the current settings in the yaml file expects the data to be present in ~/floodcase_202107/preprocessed_data, it might be good to mention this to the user.

Peter9192:
Thanks for testing Chris! I'll make all filenames consistent with the EMS poster ;-)

Peter9192:
Currently implementing the visualization (#124)

PR #131, opened 2022-09-29
Refactor time loops

Peter9192:
The original code used a nested loop like this (pseudo-code):

```py
for date in dates:
    data = load_data(date)  # hourly data, so 25 values since preproc files also included midnight of next day
    refined = interpolate('15min')  # so now 24*4+1 time steps in memory
    backtrack(refined)
    # The inner loop was hidden in the backtrack function:
    for time in refined:
        track_this_timestep()
```
As pointed out in the inline comments, this uses much more memory than is strictly needed. Moreover, the nested structure, with the inner loop hidden in a function, makes it difficult to see what's going on, and it makes it necessary to do bookkeeping in two places. Finally, the 25th hour in the input files is an ugly necessity that also complicates the preprocessing.

This PR changes the code such that it looks more like this  (pseudo-code):

```py
times = get_refined_timesteps()

for time in times:
    load_refined_data(time)  
    track_this_timestep()

def load_refined_data(time):
    upper, lower = find_bounds(time)
    load(upper)
    load(lower)
    return interp(upper, lower, t)
```

This considerably reduces memory consumption. Moreover, it removes a layer of complexity, making the code easier to understand and making it easier to e.g. write output at any desired frequency (this is now configurable).

Some side effects of this PR:
- Output writing frequency can be configured
- Nicer logging messages
- Removed output variables `s_track_upper` (and `lower`) because I felt they didn't really make sense
- Changed output to have "total" accumulations since start of tracking. This made the code much easier and it is still possible to get daily values by taking the output diff.
- I verified the output is still the same as before.

### DIFF
--- a/cases/era5_2013.yaml
+++ b/cases/era5_2013.yaml
@@ -15,7 +15,9 @@ track_start_date: '20130521'
 track_end_date: '20130604'
 
 # Settings needed for the tracking run
+input_frequency: '1h'
 target_frequency: '15min'
+output_frequency: '1d'
 periodic_boundary: false
 output_folder: ~/output_data
 restart: false
--- a/cases/era5_2021.yaml
+++ b/cases/era5_2021.yaml
@@ -15,7 +15,9 @@ track_start_date: '20210701'
 track_end_date: '20210716'
 
 # Settings needed for the tracking run
+input_frequency: '1h'
 target_frequency: '15min'
+output_frequency: '1d'
 periodic_boundary: false
 output_folder: ~/floodcase_202107/output_data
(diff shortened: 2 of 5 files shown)


Peter9192:
I can keep working on this PR, but I think it has sort of stabilized. It's now at 7 seconds per day on my laptop, still a bit slower than before but much better structured.

Peter9192:
@Imme1992 can you check if you can run the (original) 2021 flood case with this version?

Peter9192:
Thanks! Also added tagged_precip to output.

PR #139, opened 2022-10-26
Corrected mistake downloadscript

Imme1992:
I changed two things on the downloading scripts as indicated in the commits, as I discovered mean sea level pressure instead of surface pressure was downloaded, and the naming of the file was not entirely correct


Peter9192 (approved):
Thanks Imme! Well done!

PR #154, opened 2022-11-18
Download script fix

chrisweij:
Fix in download script, also made the template in the era5 yaml file consistent with the download script.


Peter9192 (changes requested):
I think we (eventually, after some back and forth) decided that we don't want to split it per day, but only per year and month.

chrisweij:
> I think we (eventually, after some back and forth) decided that we don't want to split it per day, but only per year and month.

Thanks for the remark, that should all be consistent now for downloading model and pressure levels.

Imme1992 (approved):

Peter9192 (approved):

PR #156, opened 2022-11-18
Single precision of output data

chrisweij:
As discussed, the changes in the results using single precision output are very small (on the order of 10^-4 %). Moreover the input data (at least era5) is often in the short data format, which is less precise than single precision.

### DIFF
--- a/src/wam2layers/preprocessing/era5.py
+++ b/src/wam2layers/preprocessing/era5.py
@@ -273,7 +273,7 @@ def prep_experiment(config_file):
         # Save preprocessed data
         filename = f"{date.strftime('%Y-%m-%d')}_fluxes_storages.nc"
         output_path = config["output_dir"] / filename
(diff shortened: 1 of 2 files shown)


Peter9192 (approved):
Thanks for testing @chrisweij! One remark: I think the preprocessing is currently still computing in double precision. Might be interesting to check whether we can move the "astype" conversion in the preprocessing to the data loading instead of writing step. Would you have time to check that as well? Otherwise I suggest we merge this now and test that later.

chrisweij:
> Thanks for testing @chrisweij! One remark: I think the preprocessing is currently still computing in double precision. Might be interesting to check whether we can move the "astype" conversion in the preprocessing to the data loading instead of writing step. Would you have time to check that as well? Otherwise I suggest we merge this now and test that later.

I do not have time to check this right now, but can check it some time the next weeks. So I will merge it for now.

Issue #67, opened 2022-06-24
Add logging among others about stability criterion affects fluxes (given the time frequency)


ruudvdent:

Imme1992:
Print statement or output regarding when the stability correction function is applied.
Stability criterion depends on grid spacing, time stepping, wind speeds, and where you are on the globe (grid spacing different between equator and north pole). 
Logging of criterion in separate file.

Imme1992:
Where and when criterion is not met

Imme1992:
The count of how often the stability criterion is applied (as percentage of the total grid cells) and what is its mean value per timestep is now in the main branch.

Imme1992:
Do we also want to add logging regarding the yaml file settings and copy them to the logging file which should be saved in the directory of where the backtracking is done

PR #158, opened 2022-11-18
Added logging to backtrack and profiling script

Imme1992:


Imme1992:
Pull request closes issue #67

Imme1992:
Only calculate the extra variables such as how much moisture is tracked and how often the stability criteria is used when they are also requested by the logger level in the yaml

Peter9192:
I'm afraid that currently with writing out every timestep for DEBUG the log file will be massive. Can you check this?

ruudvdent:
This is not something to merge anymore per #238

PR #163, opened 2022-11-25
Conditionally import cartopy and give a better error message if it fails

### LINKED ISSUE #149: improve warning/error cartopy availability

Peter9192:
This fixes #149 .

- Cartopy is no longer imported by default in `visualization.py`. So we can import `wam2layers.analysis.visualization`  without getting import errors.
- When any of the visualization functions is called (i.e. the users tries to make a plot), `try_import_cartopy` is called. 
  - If this works, nothing unexpected happens
  - If this fails, we get a clear error message instructing the user to install cartopy with conda.
- matplotlib and cmocean are added as standard dependencies

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -33,9 +33,9 @@ dependencies = [
     "netcdf4",
     "scipy",
     "click",
-    ## Cartopy requires GEOS and PROJ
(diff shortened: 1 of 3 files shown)


ruudvdent (approved):
Works like a charm

PR #167, opened 2022-12-02
Visualise script fix

chrisweij:
Updated visualisation script so that it now only reads the netcdf files consistent with the dates in the config file. 

Moreover, fixed the name of "s_track_upper"/"s_track_lower" to "s_track_upper_restart"/"s_track_lower_restart"


Imme1992 (approved):
Thanks, good changes!

Issue #104, opened 2022-08-18
Logo


Peter9192:
I was just wondering whether it'd be nice to have a logo, for example for use in our github organization and on my EMS poster. This is the first thing I could come up with:

![logo](https://user-images.githubusercontent.com/17080502/185397224-be6d947d-b3e1-47a2-9701-88c4d39742b9.png)

You can clearly see that I'm not a professional designer, but perhaps it helps to inspire ;-)

ruudvdent:
How about: ![Logo](https://user-images.githubusercontent.com/10234007/202754790-8b520175-bf10-46f3-baf5-6d21023536be.png)

Peter9192:
Already a big improvement!

Peter9192:
![image](https://user-images.githubusercontent.com/17080502/203531423-4bd80761-316e-41c3-b452-5ec185ccb4d8.png)

Bit more rounded version

ruudvdent:
I like it! @Peter9192 , @Imme1992 , @chrisweij . Shall we just change it to the above one in the documentation? Until we get something better of course...

chrisweij:
> I like it! @Peter9192 , @Imme1992 , @chrisweij . Shall we just change it to the above one in the documentation? Until we get something better of course...

Good for me, I like the new logo!

Peter9192:
![image](https://user-images.githubusercontent.com/17080502/205317211-0755dd2a-c0a1-46b8-9e60-baca343c6e22.png)

Peter9192:
![image](https://user-images.githubusercontent.com/17080502/209362796-2ab6ec0e-6945-4fa6-b656-efca4315950e.png)

PR #173, opened 2022-12-23
added new logo

ruudvdent:
new logo


ruudvdent:
I suggest to close #104 with this PR until we see any reason to create a new logo later

Imme1992 (approved):

docs/_static/WAM_logo_v3.png:Zone.Identifier
```diff
@@ -0,0 +1,4 @@
+[ZoneTransfer]
```
Peter9192, line 1: This file is related to WSL I believe, you can remove it. I guess it was added to git accidentily.

docs/conf.py
```diff
@@ -10,7 +10,7 @@
 
 myst_enable_extensions = ["dollarmath"]
 
-html_logo = "_static/logo.png"
+html_logo = "_static/WAM_logo_v3.png"
```
Peter9192, line 13: Old logo's are also available in git history, so we might as well keep the version out of the logo.

Peter9192 (commented):
Bit late to the party, but I noticed a few things that could be considered next time.

PR #182, opened 2023-02-06
check_for_solving_colon_error_on_Windows

ruudvdent:
The colon ":" provided problems for cloning repository on Windows . Test if this solves the problem, otherwise undo


vincentdefeiter (approved):

Peter9192:
Those `:Zone.Identifier` files are due to WSL, I think, and shouldn't be committed. Perhaps we can add the extension to `.gitignore` to avoid accidentally committing them.

Peter9192:
> So, just to be 100% clear, you are suggesting to add ".Identifier" in the .gitignore?

yes

PR #184, opened 2023-02-19
Added a source region designer python script.

vincentdefeiter:
I have added a Source Region Designer Python script under 'scripts'. I hope the branching and pull requests went correctly.


scripts/WAM2-Layers Source Region Designer.py
```diff
@@ -0,0 +1,288 @@
...
+    ax = available_regions.plot(
+        projection=ccrs.Robinson(), add_ocean=True, text_kws=text_kws
+    )
+    
+    ax.set_global()   
+    plt.show()
+    
+    
+    #create (a) mask(s) of the required region(s):
+    mask = regionmask.defined_regions.giorgi.mask_3D(longitude, latitude)   #<-- specify available regions source
```
ruudvdent, line 164: I don't get this line, because in line 110 the user could have selected another region correct? E.g. regionmask.defined_regions.ar6.land

Also it seems that `(longitude, latitude)` creates an error: `Regions' object is not callable`

scripts/WAM2-Layers Source Region Designer.py
```diff
@@ -0,0 +1,288 @@
...
+    
+    #Specify region ID - what to store
+    region_id = 2                                                           #single mask id
+    region_ids = [1]                                                        #multiple regions in one mask [id-1,id-1,...]
```
ruudvdent, line 168: are these also settings? If so, please move to settings

scripts/WAM2-Layers Source Region Designer.py
```diff
@@ -0,0 +1,288 @@
...
+    new_masks = np.nanmean(new_masks,axis=0)                                #Multiple masks into 1D array
+    
+    #Plot as a check
+    plt.contourf(longitude,latitude,new_masks)                              #check (switch between mask or masks when needed)
+    plt.show()
+    
+    #Export as source region for WAM2Layers
+    
+    # create xarray dataset
+    data = new_mask   #switch between mask or masks
```
ruudvdent, line 186: This is also a setting, correct? Please move to settings

scripts/WAM2-Layers Source Region Designer.py
```diff
@@ -0,0 +1,288 @@
...
+    
+    #create (a) mask(s) of the required region(s):
+    mask = regionmask.defined_regions.giorgi.mask_3D(longitude, latitude)   #<-- specify available regions source
+    
+    #Specify region ID - what to store
+    region_id = 2                                                           #single mask id
+    region_ids = [1]                                                        #multiple regions in one mask [id-1,id-1,...]
+    ''''!! NOTE REGION_IDS is INDEX of map - 1 !! '''
+    
+    new_mask = mask.sel(region=region_id).values
```
ruudvdent, line 171: got another error here: `'Regions' object has no attribute 'sel'`

scripts/WAM2-Layers Source Region Designer.py
```diff
@@ -0,0 +1,288 @@
...
+    plt.show()
+    
+    
+    #create (a) mask(s) of the required region(s):
+    mask = regionmask.defined_regions.giorgi.mask_3D(longitude, latitude)   #<-- specify available regions source
+    
+    #Specify region ID - what to store
+    region_id = 2                                                           #single mask id
+    region_ids = [1]                                                        #multiple regions in one mask [id-1,id-1,...]
+    ''''!! NOTE REGION_IDS is INDEX of map - 1 !! '''
```
ruudvdent, line 169: that's a bit unhandy

ruudvdent (changes requested):
Hi @vincentdefeiter, can you have a look at my comments and address them?

scripts/WAM2-Layers Source Region Designer.py
```diff
@@ -0,0 +1,288 @@
...
+output_dir = '...'
+
+#Specify latitude and longitude - same as input data to WAM2Layers
+lat_max = 80
+lat_min = -80
+
+lon_max = 180
+lon_min = -180
+
+step = 0.25 #0.25 degree intervals (ERA5 data)
```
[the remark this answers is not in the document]
vincentdefeiter: That would indeed be better, I will implement this.

scripts/WAM2-Layers Source Region Designer.py
```diff
@@ -0,0 +1,288 @@
...
+#%%
+#---------------------------------------------------------------------------
+#              S O U R C E  R E G I O N  D E S I G N E R S 
+#---------------------------------------------------------------------------
+#%%
+
+'''
+Welcome to the WAM2-Layers Source Region Designer. All methods are dependent on the 
+resolution of your ERA5 (or other) data, specified in the general settings. 
+
```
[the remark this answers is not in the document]
vincentdefeiter: I will make it more clear.

scripts/WAM2-Layers Source Region Designer.py
vincentdefeiter: I will check this.

scripts/WAM2-Layers Source Region Designer.py
vincentdefeiter: The thing with this is that it is dependent on the user. Whilst developing I noticed that it is useful to have a feedback with plotting and looking on whether you have selected the correct things. So therefore, I had intermediate plotting. Maybe, indeed as @Peter9192 suggested at the start, I could implement this more as a jupyter notebook script. In that way, you can have intermediate plotting to really see whether the thing you plot is what you want... Or, we assume that the user knows which region_id or ids to select...

scripts/WAM2-Layers Source Region Designer.py
vincentdefeiter: This can be automated better, I will implement it.

scripts/WAM2-Layers Source Region Designer.py
vincentdefeiter: It worked fine for me, I will recheck it.

scripts/WAM2-Layers Source Region Designer.py
vincentdefeiter: True, I will try to find a way to automate it, that the value really is what you select...

scripts/WAM2-Layers Source Region Designer.py
```diff
@@ -0,0 +1,288 @@
...
+    
+    > Method 3: Center point and a squared perimeter around it. 
+    
+    Here, a set center point (see settings) is set, after which a squared region (depending on the number of degrees)
+    is drawn around this center point. Take care that no emphasis is set on borders between sea/land. 
+    
+
+To set the specific method, see the settings below. Please note that you are able to specify more of the settings within the original code. 
+
```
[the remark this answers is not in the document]
vincentdefeiter: Do you want me to copy paste this license in the script?

scripts/WAM2-Layers Source Region Designer.py
```diff
@@ -0,0 +1,288 @@
...
+#%%
+#---------------------------------------------------------------------------
+#               I  M  P  O  R  T   M  O D U L E S 
+#---------------------------------------------------------------------------
+#%%
+import numpy as np
+import matplotlib.pyplot as plt
+import xarray as xr
+import cartopy.crs as ccrs
+import regionmask
```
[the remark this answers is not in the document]
Peter9192: You can add it to the dependencies section in `pyproject.tom`:

https://github.com/WAM2layers/WAM2layers/blob/422f143e391f784f6b746f7767609605e3e691fc/pyproject.toml#L27

scripts/WAM2-Layers Source Region Designer.py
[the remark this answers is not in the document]
Peter9192: MIT is very permissive. This is not a copy or a substantial portion, so don't worry about it.

scripts/WAM2-Layers Source Region Designer.py
Peter9192: Sounds like you could also have a separate function to plot/preview the region you're cropping

scripts/WAM2-Layers Source Region Designer.py
```diff
@@ -0,0 +1,288 @@
...
+METHOD 2 - Using a shapefile combined with 'geopandas' and the 'region mask' python package
+ 
+'''
+
+if selector == 2:
+    ds = gpd.read_file(shapefile)                    #load shape file
+    
+    #visualisation
+    ds.boundary.plot(color=None, edgecolor="k", linewidth=0.8)      #check
+    plt.show()
```
Peter9192, line 211: Note that plotting (showing a graphical display) is not always available, especially when you run from the command line. It might be better to save it, or make showing the plot optional

Peter9192 (commented):
Very nice work so far. Code is readable and well documented.

The main suggestion I would like to make is to structure your code with functions. Something like:

```py
def mask_with_regionmask(region_name):
    """Mask source region by using a named region from regionmask."""
    ... # here comes your code

def mask_with_shapefile(shapefile):
    """Mask source region using a shapefile."""
    ... # here comes your code

def mask_around_point(point, radius):
     """Mask source region using a center point with a given radius."""
    ... # here comes your code

if method == 1:
    mask_with_regionmask(...)
elif method == 2:
    mask_with_shapefile(...)
elif method == 3:
    mask_around_point(...)
else:
    raise ValueError("Invalid method, choose 1, 2, or 3")
``` 

Then, if we move the code to e.g. `src/wam2layers/utils/mask_region.py`, you could import it in a notebook like 

```py
from wam2layers.utils.mask_region import mask_around_point.
```

Furthermore, we could add a way to run it from the command line with a few nice options, e.g. 
```bash
wam2layers mask_region around-point --point=[52, 5] --radius=10
```
I could help you set that up once we're ready for it.

vincentdefeiter:
@Peter9192 and @ruudvdent I have committed an adjusted version of the script + example pictures. I probably did this commitment in the wrong way on GitHub, please let me know whether you can see it + whether it is okay. 

I changed it to functions + that the preprocessing files are required for fitting it to the dimensions of the data. We can still add more functionalities, e.g., combining shapefiles etc. Maybe still good to integrate it with the .yaml files in some way, but I do not know what is your standard in this.

ruudvdent (approved):

ruudvdent:
The code works fine. I only had some issues with the plotting.

1. I had to update my cartopy to version 0.21, but I suppose that's my problem rather a problem of this code
2. The preview plot only shows the world map from 0 degree (but correct netcdf-file was created nonetheless). See below.

@vincentdefeiter , could you have a final look and then merge pull request?

image

PR #196, opened 2023-04-07
Add regression tests on sample data

Peter9192:
This PR adds a very small set of test data that can be used to verify the correct functioning of WAM2layers. This would make it much easier to change the code and be sure it still works as before.


Peter9192:
Adressed in #265

PR #201, opened 2023-04-18
Movies

Peter9192:
work in progress
both_boomerang


ruudvdent:
Would be nice to eventually fully integrate in the software, but as intermediate step code could be copied/checked in notebooks folder. To do: polish the notebooks

ruudvdent:
let's archive this

PR #204, opened 2023-04-21
Check date order

Peter9192:
This PR adds a check to the config object to raise an error if the start date in the config is equal to or larger than the end date. 

It also makes sure that checks are always run, not just when the config is first loaded, but also when you try to change a value dynamically.

Finally, the PR adds tests to verify this behaviour. Running these tests as part of the PR process (by setting up [continuous integration](https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-python)) should be part of another PR


src/wam2layers/config.py
```diff
@@ -112,7 +112,7 @@ class Config(BaseModel):
 
     Should be formatted as: `"YYYY-MM-DD[T]HH:MM"`. Start date < end date, even if
     backtracking.
-    When forward tracking the track_end_date is not written out as output date. 
+    When forward tracking the track_end_date is not written out as output date.
```
Imme1992, line 115: Only a comment on forward tracking, should we say something on backward tracking here as wel?l

Imme1992 (approved):

PR #214, opened 2023-04-22
Try out issue template

Peter9192:
To see how the web interface formats them


ruudvdent (approved):
I approve so we can check it out. Will change the text later then.

PR #217, opened 2023-05-03
Improve templates

ruudvdent:
Issue template works now, just added new text.

Test if the discussion template works for the Q&A section. If yes, I'll add a template to the other sections as well.


.github/DISCUSSION_TEMPLATE/Q&A.md
```diff
@@ -0,0 +1,15 @@
...
+assignees: ''
+
+---
+
+Welcome to WAM2layers! 
+
+If you're new here, please check our [readme](https://github.com/WAM2layers/WAM2layers), 
+[documentation](https://wam2layers.readthedocs.io/en/latest/) first. 
+      
+Still questions? Please introduce yourself if you are a new user, and please ask your question as clear as possible. We hope the community gets back to you a.s.a.p.
```
Peter9192, line 15:
```suggestion
Still have questions? Please introduce yourself if you are a new user, and ask your question as clearly as possible. We try to keep an eye on this forum and we hope that everyone who benefits from it will do the same, so help can come as quickly as possible.
```

ruudvdent:
The Q&A.md doesn't seem to work, let's try the Q&A.yml again, but let's do it in parallel with two people together so we don't have to wait so long for the reviewing process.

ruudvdent:
re-opened the branch to make the discussion templates

Issue #223, opened 2023-06-08
Applying WAMlayers to cmip6 daily data


WenyuZhou-4:
Hi,
I am trying to apply this model to the daily data from the cmip6 ecearth3 model. After changing some code, the preprocessing works fine, but when I run backtracking, the final e_track results are twice as large as normal and I don't know where the problem is coming from.
I checked the input rainfall data and the e_track from the run. The total e_track in the era5 data is about 90% of the total rainfall, but in the cmip6 data it is almost 200%, which is not normal.
The Results like:
The total precipitation of era5 from 2005-2014 of whole basin in m3 is: 2.33999433e+13
The moisture evaporation of era5 from 2005-2014 in m3 is: 2.0874286e+13, which is about 89% tracked evaporation.

The total precipitation of ecearth3 from 2005-2014 in m3 is: 2.56253844e+13
The moisture evaporation of ecearth3 from 2005-2014 in m3 is: 4.697806e+13

To solve the above problem, I have made the following attempts:
1. change the targrt_frequency in config file, once it was input_freq to 1d, target_freq to 1h, now target_freq becomes 30 minutes and 10 minutes.
change the target frequency to 0.5h: 4.6247004e+13
change the target frequency to 0.1h: 4.6078267e+13
The above is the data for 10 years, in m3, which seems to have changed a bit, but not much.
2. run era5 preprocessed data with backtrackecearth3.
Result: Runs fine. This is normal, because I only modified ds.sel(time=t.to_pydatetime() + timedelta(hours=12), drop=True) on the original backtrack code, in order to apply to cmip6 daily T12:00:00 data

This result looks like forgetting to divide by 2 somewhere, but I'm not sure. Theoretically, even if the preprocessed data is completely wrong, the final result should not be greater than the total precipitation.

I have attached my modified code and provided the data for a reference. Here is the codes and datasets: 
https://drive.google.com/drive/folders/1Klg3ADH1SZ8qyqWc8YSPp7lIE-Zi2_rq?usp=sharing

Peter9192:
[...]

PR #224, opened 2023-06-12
Trying to apply WAM2layers on CMIP6 model daily data

WenyuZhou-4:
I'm trying to apply the wam2layers on cmip6 daily data. I added a new preprocessing script `ecearth3.py` that changes the names and units of the variables. Please note that the initial file still needs to be placed in the folder `/year/month/day` like in ERA5. And if you need to clip or regrid data, this should be done in a separate script.
In `shared_change.py` I made only the simplest modifications that only apply to the flux calculation for daily data, a better way is to read `input_frep` in config.
I also made changes in backtrack in order to adapt the cmip6 file. For `backtrackecearth.py` I only modified `ds.sel(time)` to work with T12:00:00in cmip6 daily data

refs #223


ruudvdent:
#223

ruudvdent:
@BSchilperoort , @Imme1992 , shall we archive this, given #394

ruudvdent:
archived, because addressed in #394

PR #226, opened 2023-06-25
Provided updated scripts for visualisation

vincentdefeiter:
I provided a list of functions to make more detailed visualisations than the standard visualisation functions. 

The following functions are prescribed: 

- Logfile. This function provides a logfile (.txt) with the evaporation recycling ratio (evaporated water that returns as precipitation), pecipitation recycling ratio (region�s dependence on evaporation from within the region for precipitation), sum of the precipitation, evaporation and their respective absolute differnce. This distance should be kept to a minimum and can be used as a check whether the model is operated correctly. 

- Precipitation. This function develops a figure showing the precipitation events tracked by WAM2Layers. It visualises the source region accompanied by the precipitation. 

- Evaporation. This function develops a figure showing the tracked evaporation (showing source and sinks) from the source region, with the accompanying moisture fluxes (arrows) and mean surface pressure (contours in hPa).

- Moisture_fluxes. This function develops figures showing the mean moisture fluxes (arrows) with their corresponding magnitude (contours). It develops a figure for a specific time range and for both the lower and upper layer of the model (seperated by the p_divide). 

Please have a look and see whether it works on your model in- and output. I cannot yet provide examples since they are based on my Thesis results. I will update this pull request once I have different examples.


chrisweij:
Thanks Vincent, great work! I got it working again with the current version of WAM2layers and I started to clean up the code a bit.

- Logfile: I think it would be most useful to calculate some area-averaged instead of a sum, since the sum is hard to interpret and depends also on the area size. 
- Precipitation: I moved the small inset showing the region on a larger scale map a bit so that it did not overlap with the x-axes ticks. Otherwise I think it looks great, maybe we should include an explicit option to choose a colour map. 
- Evaporation: u, and v are read in, but I do not think they are used. Pressure on more contour levels? 
- Moisture_fluxes: u and v are read in, but are not used. Moreover, it might be nice to also have an option to create an animation of the change of moisture fluxes over a certain time period, instead of only the mean fluxes. 

Lastly, there is still code at the bottom of the script which probably fits better in just the python notebook.

ruudvdent:
@chrisweij , is this ready for review with your latest changes?

chrisweij:
@ruudvdent Not completely yet, still have to clean up the logging text a little bit, will do this asap.

chrisweij:
Should be almost ready for review. While looking at the logging results, I found that the absolute amount of moisture sources (in kg) is significantly higher than the tagged precipitation (in kg). So I might still make an error in the calculation, or some thinking error, or a file is missing. 

As an alternative we can already use what we already have here, and then see if we can solve this issue later.

ruudvdent:
> Should be almost ready for review. While looking at the logging results, I found that the absolute amount of moisture sources (in kg) is significantly higher than the tagged precipitation (in kg). So I might still make an error in the calculation, or some thinking error, or a file is missing.
> 
> As an alternative we can already use what we already have here, and then see if we can solve this issue later.

I'll have a look asap

ruudvdent:
As discussed today. We will move this to a new repository WAM2layers/UserExamples

ruudvdent:
Moved here: https://github.com/WAM2layers/user-examples

PR #231, opened 2023-07-17
Fix output filenames to include hours and minutes

### LINKED ISSUE #229: Fixing time stepping frequency output
Issue with output filenames not including hours and minutes (and therefore always daily)

Peter9192:
This PR solves a bug where the output filenames only include dates and no time. When the output frequency is set to more than one output file per day, the previous file is overridden. This PR adds hours and minutes to the output files.

Closes #229 
Refs #218


Imme1992 (approved):

Peter9192:
Thanks for the review

PR #238, opened 2023-07-18
Logging

### LINKED ISSUE #67: Add logging among others about stability criterion affects fluxes (given the time frequency)

Peter9192:
Supersedes #158 

Replaces print statement by calls to logging.info
Write logging.info to terminal and logging.debug to logfile (`wam2layers_<timestamp>.log`)

Should close #67


Peter9192:
@Imme1992 @ruudvdent in #158 we added prints of all the stability corrections. However this means there's 12 prints per tracked hour (at a time step of 10 minutes). That leads to enormous log files. Can we come up with a smarter approach? For example, only log when the correction exceeds a certain amount?

Peter9192:
After discussion, we settled on the following

1. First time a stability correction is applied for more than 5% of the grid, issue a warning advising to reduce the timestep
2. Subsequently, once the previous correction is exceeded by either 5%point of grid cells, or doubled in magnitude, issue the warning again.

Peter9192:
@Imme1992 @ruudvdent should we track and log the `mean` stability correction of all grid cells or the `maximum`?

ruudvdent:
> @Imme1992 @ruudvdent should we track and log the `mean` stability correction of all grid cells or the `maximum`?

Could we do both? Information could be relevant for different purposes. 

> First time a stability correction is applied for more than 5% of the grid, issue a warning advising to reduce the timestep
    Subsequently, once the previous correction is exceeded by either 5%point of grid cells, or doubled in magnitude, issue the warning again.

Would be useful to for those specific timesteps write out a netcdf-file containing that correction, so I user could inspect at which locations a stability correction is applied.

Peter9192:
> > First time a stability correction is applied for more than 5% of the grid, issue a warning advising to reduce the timestep
> > Subsequently, once the previous correction is exceeded by either 5%point of grid cells, or doubled in magnitude, issue the warning again.
> 
> Would be useful to for those specific timesteps write out a netcdf-file containing that correction, so I user could inspect at which locations a stability correction is applied.

I like that idea, we could also do a dump of the full state or something like that. Does require quite a bit of refactoring though

Imme1992:
I am looking into the logging branch, it looks good, my suggestion so far is to copy the log file to the output folder instead of having it in the working directory folder.

Imme1992 (commented):
See my comment in the logging

Peter9192:
Thanks for reviewing! Filename of the correction field is fixed now, and I created https://github.com/WAM2layers/WAM2layers/issues/267 about changing the location of the logfile. Let's pick that up in a follow-up PR

Imme1992 (approved):

PR #241, opened 2023-07-19
Adjusted line 178 for equal pressure levels to >=

vincentdefeiter:
I adjusted the line 178 in era5 pre-processing. I ran into the problem of having dp = 0 due to pressure level and surface pressure being exactly equal. By changing > to >=, this problem seems to be resolved. By comparing pressure and model levels, we can see whether this results in any issues.

### DIFF
--- a/src/wam2layers/preprocessing/era5.py
+++ b/src/wam2layers/preprocessing/era5.py
@@ -171,8 +171,6 @@ def get_dp_pressurelevels(q, u, v, ps, qs, us, vs):
     q = q.assign_coords(level=np.arange(nlev))
     p = p.assign_coords(level=np.arange(nlev))
 
-    # import IPython; IPython.embed(); quit()
-
     # Calculate pressure jump
     dp = p.diff("level")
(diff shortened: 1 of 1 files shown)


src/wam2layers/preprocessing/era5.py
```diff
@@ -171,9 +171,11 @@ def get_dp_pressurelevels(q, u, v, ps, qs, us, vs):
     q = q.assign_coords(level=np.arange(nlev))
     p = p.assign_coords(level=np.arange(nlev))
 
+    # import IPython; IPython.embed(); quit()
```
ruudvdent, line 174: can you delete this line?

ruudvdent (changes requested):

ruudvdent (approved):

ruudvdent:
@vincentdefeiter will merge this once his plots show that there indeed is no further issue

ruudvdent:
Let's update this anyway now, because this branch will start the run too far behind.

PR #249, opened 2023-07-21
copy config yaml file to output folder

### LINKED ISSUE #267: Session handling, logfiles & config files
Currently (after #238) the logfile is written to the current working directory. Instead, it would be helpful if the logfile was written to the session output directory. However, the cli runner that creates the logfile doesn't know about the output dir, and the library code that knows the output dir doesn't know about the logging handler. 

Perhaps we can use click's contexts handling to implement this neatly. https://click.palletsprojects.com/en/8.1.x/commands/#nested-handling-and-contexts

Imme1992:
Copy config yaml file to output folder for all command line executions.

This PR closes #267.


ruudvdent:
check again after merging PR #238

Peter9192 (approved):
Nice work!

PR #250, opened 2023-08-11
modified lower and upper tracking to ensure that the maximum fraction tracked never exceeds 100%

eabarnes1010:
This 3-line edit in the backtrack.py fixes a bug where the amount tracked could exceed 100%. While in many cases this was not a proble, there were certain locations and timestamps where this caused the backtracking to explode (i.e. percent tracked exceeding 1000000%, etc). The fix is to cap the fraction tracked to 1.0.

### DIFF
--- a/src/wam2layers/tracking/backtrack.py
+++ b/src/wam2layers/tracking/backtrack.py
@@ -323,8 +323,9 @@ def backtrack(
     ) = to_edges_meridional(fy_upper)
 
     # Short name for often used expressions
-    s_track_relative_lower = s_track_lower / s_lower
-    s_track_relative_upper = s_track_upper / s_upper
+    s_track_relative_lower = np.minimum(s_track_lower / s_lower, 1.0)
+    s_track_relative_upper = np.minimum(s_track_upper / s_upper, 1.0)
+
     if config.periodic_boundary:
         inner = np.s_[1:-1, :]
     else:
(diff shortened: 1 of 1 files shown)


Peter9192:
Hi @eabarnes1010, congrats on opening your first PR in this repo! 

Did you know it is possible to [link PRs to issues](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue)? For example, I think this one closes #230. It is common practice to put such links in the top post (you can still edit that). This will provide additional context for reviewers and others that stumble upon this PR later. Also, the "magic" word `closes` will automatically close the issue once this PR is merged.

I think you shared some additional information before via email on how you diagnosed the problem. It might be nice to add that to the corresponding issue or in this PR, also for context.

As you can see, some checks were automatically running once you pushed. Note that these only check if the model still runs - they don't perform any validation of output at the moment. Nevertheless, it seems you didn't break the model 👍. I'll now do a formal review for illustration purpose 😉

Peter9192:
By the way, when you're doing working on the PR and all tests pass, you can mark it as "ready for review" (and potentially request a review from someone)

eabarnes1010:
I looked for a while and nothing jumped out at me as a big logical flaw anywhere. I think this comes about just because ERA5 does not conserve water, so small amounts appear and reappear from time to time. In this case, the issue is over a high elevation region of the Andes which generally makes sense to me. So I think that this fix is all that is needed - we have now successfully backtracked the entire globe for a decade with no issues after this fix was added.

Peter9192 (approved):
Okay, based on your response I'll approve this one. It would be good to think about potentially logging a warning when this correction is applied, just like in https://github.com/WAM2layers/WAM2layers/pull/238. But we can do it later.

Peter9192:
Thanks for the bugfix! It's now merged 🎉  Note that it will require a new release before these changes are available when you install from PyPI.

PR #251, opened 2023-08-11
select a subregion for backtracking in the config file

eabarnes1010:
added option for user to specify (in the config file) the domain over which the backtracking is done. this code also allows to flexibility in the domain if the reduce domain crosses the prime meridian. Important note is that the source file (nc file) must be saved at the same resolution as the raw ERA5 data (e.g. the globe; not the reduced domain)...otherwise the code is slower.

This PR refs #252.


Peter9192:
I know you're still working on this one, but just upfront: normally it's a good idea to open an issue first to discuss the idea. In this case, we need to think carefully about a few things, for example what should be the order of the lat/lon bounding box, or whether this only affects the tracking or also pre-processing. So far the way to do this had been to make a reduced "src_region" file. What happens if these options are mixed?

eabarnes1010:
Ruud asked me to do this. It sounds like it might be more work that it is worth if your group won't use it. We can always just use it here then.

Peter9192:
Okay, understood! I still think it might be useful, so what I suggest is to use this draft PR as a "proposal" that we can subsequently discuss if/how to proceed.

eabarnes1010:
Ok - put an issue in to start discussion!
-Libby

---------
Dr. Elizabeth A. Barnes | she, her, hers
Professor | Dept. of Atmospheric Science | Colorado State University
Email: ***@***.***
Google: ***@***.***
Website: *http://barnes.atmos.colostate.edu/
<http://barnes.atmos.colostate.edu/>*
Scheduling: *https://calendly.com/eabarnes <https://calendly.com/eabarnes>*

src/wam2layers/tracking/backtrack.py
```diff
@@ -10,6 +10,66 @@
...
+    # get the config domain bounds specified by the user
+    bounds = config.reduced_domain_bounds  # [south_lat, north_lat, left_lon, right_lon]
+    south_lat, north_lat = bounds[0], bounds[1]
+    left_lon, right_lon = bounds[2], bounds[3]
+
+    # check if need to map longitude to -180 to 180 for better definition of the domain
+    if left_lon > right_lon:
+        da.coords["longitude"] = (da.coords["longitude"] + 180) % 360 - 180
+        da = da.sortby(da.longitude)
+        da = da.transpose(..., "latitude", "longitude")
```
Peter9192, line 25: Do you remember why that transpose was needed?

src/wam2layers/tracking/backtrack.py
```diff
@@ -10,6 +10,66 @@
...
+        right_lon = (right_lon + 180.) % 360. - 180.
+        assert left_lon < right_lon, "Something is not right. Try flipping the order of your longitude bounds."
+
+        check_periodic_boundaries(longitude_units="180_180", config=config)
+    else:
+        check_periodic_boundaries(longitude_units="0_360", config=config)
+
+    mask_lon = (da.longitude >= left_lon) & (da.longitude <= right_lon)
+    mask_lat = (da.latitude >= south_lat) & (da.latitude <= north_lat)
+    return da.where(mask_lon & mask_lat, drop=True)
```
Peter9192, line 37: I'm wondering if you could alternatively do 
```suggestion
    return da.sel(longitude=Slice(left_lon, right_lon), latitude=Slice(south_lat, north_lat))
```
I think that should also work for a regular lat/lon grid (haven't tested though), and I think it could be more memory efficient.

src/wam2layers/tracking/backtrack.py
```diff
@@ -43,7 +103,8 @@ def output_path(date, config):
 def read_data_at_date(d):
     """Load input data for given date."""
     file = input_path(d, config)
-    return xr.open_dataset(file, cache=False)
+    da = xr.open_dataset(file, cache=False)
+    return reduce_domain(da, config)
```
Peter9192, line 107: This is the best place to do it for now, I think. The advantage is that you never read the redundant data into memory.  If we ever switch to time-dependent tracking_regions (specified through .nc file) we might need to move it around a bit.

src/wam2layers/tracking/backtrack.py
```diff
@@ -470,6 +532,11 @@ def run_experiment(config_file):
         fluxes = load_fluxes(t)
         states_prev, states_next = load_states(t)
 
+        # Reduce the domain
+        fluxes = reduce_domain(fluxes, config)
+        states_prev = reduce_domain(states_prev, config)
+        states_next = reduce_domain(states_next, config)
```
Peter9192, line 538: If reduce_domain is already called on `read_data`, why is it still needed here?

Peter9192 (commented):
From a quick first look, I have a few questions about this code so far. Would be great if you could clarify. Thanks

src/wam2layers/tracking/backtrack.py
eabarnes1010: Yes, because the sort by moves the order of latitude and longitude around in the way the array is stored, and later in the WAM code these arrays are called via slicing (e.g. line 165) so latitude and longitude must stay in the consistent order.

src/wam2layers/tracking/backtrack.py
eabarnes1010: I think this switch is good. Slicing in this way is inclusive I believe, so should repeat my code exactly. (this replacement won't work if it is an OR, instead of an AND...but that is not relevant for this code snippet).

PR #258, opened 2023-09-03
Updated visualisation of source region designer

vincentdefeiter:


ruudvdent (changes requested):
@vincentdefeiter 

Tested and works. However:

- it's nicer if people can choose their own file name (so one could e.g. create 20 different source regions in a loop). So as it was before. The most important thing for now is that the variable name in the netcdf-file is 'source_region'. Please fix with prio in this PR.
- the preview visualization still provides a weird view when choosing regions around the Greenwich line. I'll post a result below. Can be separate issue, does not need to be fixed in this PR.

ruudvdent:
code:
```mask_with_regionmask('prudence',[1],input_file, output_dir, return_plot=True)```

image

vincentdefeiter:
@ruudvdent I have made the requested changes as a new commit. I see that some package problems arise. Let's discuss. Regarding the visualisation options, this is, I believe, something beyond my control.

ruudvdent:
seems to work fine:

image

ruudvdent (approved):

PR #259, opened 2023-09-14
Start to generalize some functions

Peter9192:
This PR contains some exploratory work that I've been doing. 

- Move data loading / writing to a generalized `io.py` module
- Create a `time.py` module for standardized time loops
- Mostly unrelated: see what a BMI interface for WAM2layers might look like (with forward tracking)

In the end, I want to see if we can reuse the interpolating/sliding window time loop also for the pre-processing.


Peter9192:
Mostly superseded by #278

PR #271, opened 2023-10-05
Test output of visualize workflow

geek-yang:
In this PR we add an extra test for the output of visualize workflow:

- Test of visualize workflow by verifying the output image
- Setup conda environment in ci to support `cartopy`

In this PR, we also fix the broken ci due to the logger.


tests/test_workflow.py
```diff
@@ -40,3 +41,22 @@ def test_backtrack():
...
+    output_path = Path("tests/tmp/output_data/figures/cumulative_sources.png")
+    assert output_path.exists()
+    expected_output = Path("tests/test_data/verify_output/cumulative_sources.png")
+    stdout = matplotlib.testing.compare.compare_images(
+        output_path, expected_output, tol=5
+    )
+    if stdout:
+        raise ValueError(
+            "Output figures are different! Please check the difference in the produced"
+            " `diff` figure and verify your results."
```
Peter9192, line 61: Can you please use the full paths to the diff file? And also it would be really nice to say something like "If you want to keep the new results, you can update the reference figure with `cp tests/tmp/output_data/figures/cumulative_sources.png tests/test_data/verify_output/cumulative_sources.png` and commit the changes.

Peter9192 (approved):
Nice work! One suggestion :-)

tests/test_workflow.py
geek-yang: Nice suggestion! Implemented!

PR #274, opened 2023-10-05
Update vertical flux calculations

### LINKED ISSUE #143: Check the vertical fluxes
Is it possible to use the vertical flux from the model to determine the vertical moisture flux?
Test how that affects the results
Test for winter (synoptic situation case) vs summer case (convection)

### LINKED ISSUE #263: Refactor shared parts
Currently some parts of the code like I/O and configuration are reused in several scripts (pre-processing, tracking, post-processing), and they are either re-implemented or imported. It would be nice to standardize this a bit more so that it's consistent and can also be used when you e.g. interactively analyze the output in a notebook. I made a start with this in #259

Peter9192:
For total moisture
* Fixes a bug (?!) in the calculation of Fz, effectively changing the direction of the vertical flux.
  Previous versions of WAM2layers used equation 2.8 from [Van der Ent 2014](https://repository.tudelft.nl/islandora/object/uuid%3A0ab824ee-6956-4cc3-b530-3245ab4f32be), this PR effectively multiplies that equation with `-1`.
* Rename "previous" and next to be less confusing
* Verify that the derivation of Fv is correct
* Use the *new* state in the error correction term `S_1 / S_T * err_T` 
* Add additional documentation and doctest for vertical advection terms

For tagged moisture:
* Seperates the vertical advection and dispersion terms: dispersion is modelled as `kvf * abs(Fv) * dS/dz`

Furthermore:
* Splits the backtracking module, moving functions to separate `io`, `core`, and `preprocessing.shared` modules.

closes #143
closes #263


Peter9192:
Using the brand new test function to check that I'm not breaking stuff. So far, after what *should* be only some refactoring, I get the following differences:

```
FAILED tests/test_workflow.py::test_backtrack - AssertionError: 
Not equal to tolerance rtol=1e-07, atol=0

Mismatched elements: 290 / 1681 (17.3%)
Max absolute difference: 0.00097656
Max relative difference: 1.3000457e-06
 x: array([[0.000000e+00, 0.000000e+00, 0.000000e+00, ..., 2.407820e-12,
        8.363298e-13, 0.000000e+00],
       [0.000000e+00, 0.000000e+00, 0.000000e+00, ..., 6.730715e-15,...
 y: array([[0.000000e+00, 0.000000e+00, 0.000000e+00, ..., 2.407820e-12,
        8.363298e-13, 0.000000e+00],
       [0.000000e+00, 0.000000e+00, 0.000000e+00, ..., 6.730714e-15,...
=================== 1 failed, 1 passed, 1 warning in 13.80s ====================
```

That seems like a very small difference indeed, and could simply be round-off errors. For reference, when I change the sign of fv I get a max relative differences of 0.67, which is much larger indeed.

Peter9192:
Here's a snapshot of the difference field:

image

Looks quite random to me.

Peter9192:
The latest commit changed which states were used in the calculation of Fz. Before it was the 'mean' state between t and t+1, now it is the state after the update. The difference due to this fix is in the order of 1%:

```
FAILED tests/test_workflow.py::test_backtrack - AssertionError: 
Not equal to tolerance rtol=1e-07, atol=0

Mismatched elements: 341 / 1681 (20.3%)
Max absolute difference: 0.00097656
Max relative difference: 1.3000457e-06
 x: array([[0.000000e+00, 0.000000e+00, 0.000000e+00, ..., 2.407820e-12,
        8.363298e-13, 0.000000e+00],
       [0.000000e+00, 0.000000e+00, 0.000000e+00, ..., 6.730715e-15,...
 y: array([[0.000000e+00, 0.000000e+00, 0.000000e+00, ..., 2.407820e-12,
        8.363298e-13, 0.000000e+00],
       [0.000000e+00, 0.000000e+00, 0.000000e+00, ..., 6.730714e-15,...
=================== 1 failed, 1 passed, 1 warning in 16.84s ====================
```

Peter9192:
The latest commit fixed a bug (?!), effectively reversing the direction of the vertical flux. Our regression tests now report a 0.5 maximum relative difference. The results of our example floodcase seem to be quite similar still. Here's a difference plot of the regression test case:

image

src/wam2layers/utils/profiling.py
ruudvdent, line 1: @Peter9192 , what happened to this profiling now then?

src/wam2layers/preprocessing/era5.py
ruudvdent, line 1: @Peter9192 , what's the reason behind this change?

ruudvdent (commented):
In the process if performing a testrun myself, however, I can already suggest to add the substantial changes to the changelog

ruudvdent:
closes #143

ruudvdent (commented):
Everything works although I believe it's rather slow now? 35 seconds per day for the floodcase, with 10 min timestep, didn't it use to be faster?

ruudvdent:
cumulative_sources

Looks reasonable to me

ruudvdent (approved):
Approved, but @Peter9192 , could you leave some comments on the questions I had during the review?

src/wam2layers/preprocessing/era5.py
Peter9192: Looks like removal of dead code. I guess `isort` detected that these imports were unused.

src/wam2layers/utils/profiling.py
Peter9192: It's still there, but I've changed the code such that the east/west loss etc are now accounted in e_track, so no need to add them separately anymore. Removing lines 35 and 50-55 are the only changes here.

Peter9192:
> Everything works although I believe it's rather slow now? 35 seconds per day for the floodcase, with 10 min timestep, didn't it use to be faster?

You're right, in my tests its 17 seconds now vs 10 seconds with the current main branch. Might be that we're doing some extra work now in the staggering of the grid, for example. The upside is that it's now easier to profile the code and pinpoint where we're spending a lot of time. I might also have removed some of the caching of input data, perhaps we can bring it back at some point.

Peter9192:
Thanks for the review!

PR #291, opened 2023-12-18
Move log files to output path

geek-yang:
Move log files to the output path instead of the current working directory.

The "setup_log" function is now executed in each child `click` command instead of in the parent `click` group.

Note that, in order to make sure the `.log` file will be created before the getting logger behavior, `logger = logging.getLogger(__name__)`, we need to move the `click` functions to the top of each file.

This PR refs #267 .


geek-yang:
Copying `config_file.yaml` file to output path is currently a working progress in another PR #249. I think it is better that we get this PR merged and then deal with the copying of config file.

Peter9192:
> Note that, in order to make sure the `.log` file will be created before the getting logger behavior, `logger = logging.getLogger(__name__)`, we need to move the `click` functions to the top of each file.

I don't think that should be the case. Notice that mostly everything in each file is "library code". Library code should just spawn log messages as the library developer sees fit. Doing something with those log messages is the responsibility of the "application", i.e. the cli programme defined with click. If anything, they should be less entangled, not more.

geek-yang:
> > Note that, in order to make sure the `.log` file will be created before the getting logger behavior, `logger = logging.getLogger(__name__)`, we need to move the `click` functions to the top of each file.
> 
> I don't think that should be the case. Notice that mostly everything in each file is "library code". Library code should just spawn log messages as the library developer sees fit. Doing something with those log messages is the responsibility of the "application", i.e. the cli programme defined with click. If anything, they should be less entangled, not more.
> 
> See e.g. https://stackoverflow.com/a/27017068

Nice suggestion! Following this, I move all command line codes to one file and now they are separated from the library code, which looks way much cleaner! 😄

src/wam2layers/analysis/visualization.py
```diff
@@ -120,6 +119,7 @@ def visualize_input_data(config_file):
     out_dir = Path(config.output_folder) / "figures"
     out_dir.mkdir(exist_ok=True, parents=True)
     fig.savefig(out_dir / "input_event.png", dpi=200)
+    logger.info("Generate figure showing the cumulative moisture inputs.")
```
Peter9192, line 122:
```suggestion
    logger.info("Generated figure showing the cumulative moisture inputs.")
```
It could be nice to also include the output path, e.g. "Figure of cumulative outputs written to `path/to/figure`"

src/wam2layers/cli.py
```diff
@@ -1,15 +1,19 @@
 """Command-line interface for the WAM2layers model."""
```
Peter9192, line 1:
```suggestion
"""Command-line interface for the WAM2layers model.

This module contains all the functionality that makes it possible to run wam2layers on the command line, like so:

    wam2layers preprocess era5 floodcase.yaml
    wam2layers backtrack floodcase.yaml

et cetera. It is built with [click](https://click.palletsprojects.com/en/8.1.x/).
"""
```

src/wam2layers/tracking/backtrack.py
```diff
@@ -203,33 +202,3 @@ def run_experiment(config_file):
...
-    CONFIG_FILE: Path to WAM2layers experiment configuration file.
-
-    Usage examples:
-
-        \b
-        - python path/to/backtrack.py path/to/cases/era5_2021.yaml
-        - wam2layers backtrack path/to/cases/era5_2021.yaml
-    """
-    logger.info("Welcome to WAM2layers.")
-    logger.info("Starting backtrack experiment.")
```
Peter9192, line 230: I liked this welcome message. Can we keep it and make this consistent also for the other CLI commands?

Peter9192 (approved):
I like this! It triggers a whole new stream of ideas about further improvements. For now, I have just a few small suggestions.

src/wam2layers/tracking/backtrack.py
geek-yang: Yes, now we have this welcome message in all cli commands.

src/wam2layers/analysis/visualization.py
geek-yang: Cool! The path information is added to the log info.

PR #292, opened 2023-12-19
Recommend python 3.11 in installation docs

Peter9192:
This is generally faster and fixes hdf5 issues with dask


geek-yang (approved):
Looks cool!

PR #293, opened 2023-12-19
Preprocess one time step at a time

Peter9192:
By manually stepping over each time step in the output data, we can preprocess everything without using dask. This makes it easier to do logging et cetera, and we don't have the overhead of dask.

A potential downside of it is that with dask, you could have pre-processed, let's say, 7 timesteps at once if the memory permitted it. And it could have done it in parallel. Now, we're bound to the limit of 1 timestep at a time, one after the other.

Since xarray can't append along a dimension, I have used a snippet from one of their github issues: https://github.com/pydata/xarray/issues/1672#issuecomment-685222909

TODO: 

- [x] Now I'm appending time steps to an existing file, instead of (over)writing an entire file at once. Due to this, if the file already exists, things get messy. It would be good to have a better handling of "overwriting" behaviour.
- [x] Simplify the code snippet a bit that I copied from the xarray gh thread (only needs to deal with time, always one step at a time)

### DIFF
--- a/src/wam2layers/config.py
+++ b/src/wam2layers/config.py
@@ -271,31 +271,6 @@ class Config(BaseModel):
 
     """
 
-    chunks: Optional[Dict[str, int]]
-    """Whether to use dask.
-
-    Using dask can help process large datasets without memory issues, but its
-    performance is quite sensitive to the chunk configuration.
-
-    Some examples:
-
-    .. code-block:: yaml
-
-        # don't use dask:
-        chunks: null
-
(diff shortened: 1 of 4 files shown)


Imme1992:
Thanks for the changes @Peter9192 ! The values of the variables in the updated preprocessing did not change, so good to go in my opinion given the two comments that I included

Imme1992 (changes requested):

Imme1992 (approved):

PR #294, opened 2023-12-20
Adjusted source region into tagging region

vincentdefeiter:
- Renamed source region files and notation into 'tagging region'. 
- Created .nc-file now has 'tagging_region' as data variable. Note, this should be adjusted for the full code. 
- I added an additional plot if you use a shapefile to develop a tagging region. You now get a zoomed-in and zoomed-out plot if you state return_plot = True. The additional plot will aid in verifying whether the 'tagging region' has been selected appropriately (e.g., you see shapes of continents etc.).


Peter9192 (approved):
Thanks for renaming things and figuring out the tests :-)

PR #297, opened 2023-12-20
Harmonizing regions for tagging and tracking

### LINKED ISSUE #55: let backtracking script check for 2D/3D mask and for values being between 0 and 1

### LINKED ISSUE #252: need for subsetting regions for backtracking
As it stands, the backtracking code will run over the entire domain of the era5 data. However, one should not have to re-download (or resave) the ERA5 data for each separate domain they may want to perform the backtracking on. In my view, the best way to use the WAM for future flexibility is to download the global ERA5 data and run the pre-processing (fluxes/states) on that data. Then, at the backtracking stage, have the user specify the domain over which to perform the backtracking. This could be a subset of the globe, or the entire globe. I have written code to do this, but am putting in an issue to start discussion.

Peter9192:
Supersedes #251 
Closes #252 
Closes #55 

After this PR, you can 

- Specify a `tracking_domain` in the config. The preprocessed data will be cropped to this domain.
- Specify a `tagging_region` (replaces `region`) as a file (netcdf with ones and zeros) or a bbox.
- If the tagging region is a file, it can have a time dimension as well

To do:
- [x] Update config
- [x] Update example config
- [x] Fix test for config file (use test file for source region so path exists)
- [x] Add validators for bbox
- [x] Select subdomain for simple cases (fits within [-180, 180])
- [x] Select subdomain for complex cases (does not fit within [-180, 180])
- [x] Use path with 2d netcdf + start/end time for tagging_region  
- [x] Use path with 3d netcdf + start/end time for tagging_region  
- [x] Use bbox + start/end time for tagging_region  
- [x] Also apply changes to forward tracking


Peter9192:
This could still be polished a bit and might need more thorough testing, but the basis is ready, I think.

geek-yang:
Now we can also update the ci to let it run all tests.

src/wam2layers/config.py
```diff
@@ -49,16 +80,45 @@ class Config(BaseModel):
...
+    """Subdomain delimiting the region considered during tracking.
+
+    This is useful when you have global pre-processed data but you don't need
+    global tracking.
+
+    You can specify a bounding box of the form [west, south, east, north].
+
+    The bounding box should be inside -180, -80, 180, 80; if west > south, the
+    coordinates will be rolled to retain a continous longitude.
+
```
geek-yang, line 116:
```suggestion

If it is set to `null`, then it will use full domain of preprocessed data.

```

Let user know this option from documentation.

ruudvdent:
Idea to use terminology "tagging_region" and "tracking_domain"

src/wam2layers/analysis/visualization.py
```diff
@@ -67,13 +67,13 @@ def _plot_precip(config, ax):
 
 def _plot_evap(config, ax):
     """Return subplot with tracked evaporation."""
-    region = load_region(config)
+    region = load_tagging_region(config)
```
geek-yang, line 70:
```suggestion
    region = load_tracking_region(config)
```

Right?

Peter9192: No, I think this used to plot only the domain for which is tagged. But it might not make sense with all the new options.

example_config.yaml
```diff
@@ -1,22 +1,28 @@
...
 level_type: model_levels
 levels: [20,40,60,80,90,95,100,105,110,115,120,123,125,128,130,131,132,133,134,135,136,137]
 
 # Tracking
-region: /data/volume_2/era5_2021/source_region_global.nc
-track_start_date: "2021-07-01T00:00"
-track_end_date: "2021-07-15T00:00"
-event_start_date: "2021-07-13T00:00"
-event_end_date: "2021-07-15T00:00"
+tracking_region: null  # use full domain of preprocessed data
```
Imme1992: tracking_region should become tracking_domain

example_config.yaml
```diff
@@ -1,22 +1,28 @@
...
+# tracking_region: [0, 50, 10, 55]  # use subdomain of preprocessed data
```
Imme1992: tracking_region should be tracking_domain

Imme1992 (changes requested):

example_config.yaml
```diff
@@ -1,22 +1,28 @@
 # General
 log_level: info
-preprocessed_data_folder: ~/floodcase_202107/preprocessed_data
-output_folder: ~/floodcase_202107/output_data
+preprocessed_data_folder: tests/tmp/preprocessed_data
+output_folder: tests/tmp/output_data
 chunks: null
```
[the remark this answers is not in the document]
Peter9192: Done, thanks!

chrisweij:
The 3d tagging can be read, however one issue is that it now continues after (i.e. earlier in time) with the last provided location. Either the tagging begin_date or the time stamp in the tagging region netcdf should be used.

example_config.yaml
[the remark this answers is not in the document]
Peter9192: Good point, let's make an issue about that to add it later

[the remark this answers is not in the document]
Peter9192: Added a note in the documentation

example_config.yaml
```diff
@@ -1,22 +1,28 @@
...
-event_start_date: "2021-07-13T00:00"
-event_end_date: "2021-07-15T00:00"
+tracking_region: null  # use full domain of preprocessed data
+# tracking_region: [0, 50, 10, 55]  # use subdomain of preprocessed data
+tracking_start_date: "2021-07-01T00:00"
+tracking_end_date: "2021-07-15T00:00"
+
+# Tagging
+tagging_region: tests/test_data/test_region.nc
+# tagging_region: [0, 50, 10, 55]  # alternative to mask file
```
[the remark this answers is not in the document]
Peter9192: Done

example_config.yaml
Peter9192: Done

example_config.yaml
Peter9192: Done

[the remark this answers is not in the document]
Peter9192: Added a warning

chrisweij (approved):

Imme1992 (approved):

PR #299, opened 2023-12-20
Moving tagging region

chrisweij:
Still in conflict because of changing names to target_region

New function takes as input a list of points of the form as first input:
[(41.5, -58.0),
 (41.5, -57.0),
 (41.5, -57.0),
 (41.5, -56.5),
 (41.5, -56.0),
 (41.0, -54.5),
 (41.0, -54.0),
 (41.0, -54.0),
 (41.0, -53.0),
 (40.5, -51.5),
 (40.5, -51.5),
 (40.5, -50.5),
 (40.5, -50.0),.....]


Peter9192 (approved):
Thanks, nice that we can do this now!

PR #303, opened 2023-12-21
Bug fix in forwardtrack script

Imme1992:
Changed minimum to maximum and handling data array with values


Peter9192 (approved):
I've seen that it works on your end.

Peter9192:
@ruudvdent notice this as well.

Issue #300, opened 2023-12-21
Fix stability for negative values in S_track


ruudvdent:
We encountered rare negative values in S_track during a forward tracking, however, this is likely not only relevant for forward, but also for backward tracking. The problem was likely introduced by applying the tagged water balance in one go rather than for advection, vertical and surface fluxes succesfully as in the WAM2layers-v2 version.

ruudvdent:
fixed in #305

PR #305, opened 2023-12-22
Losses and gains

### LINKED ISSUE #295: Treatment of losses
Currently boundary losses are added to `e_track` or `p_track`. It would be nicer to add another field to the output file called losses. This field could also include losses in the interior due to inaccuracies (numerical or due to discontinuitities in input data).

ruudvdent:
bookkeeping of losses and gains. closes #295 #300


ruudvdent:
closes # 295 #300

src/wam2layers/tracking/backtrack.py
```diff
@@ -61,6 +61,15 @@ def backtrack(
         + region * precip * s_lower / (s_upper + s_lower)
         - evap * s_track_relative_lower
     )
+    # TODO: find better way to deal with negative values
```
Imme1992, line 64: remove line 64 as we have addressed this now?

src/wam2layers/tracking/backtrack.py
```diff
@@ -70,31 +79,47 @@ def backtrack(
         )
         + region * precip * s_upper / (s_upper + s_lower)
     )
+    # TODO: find better way to deal with negative values
```
Imme1992, line 82: Remove line 82 as we have addressed this now

ruudvdent: done

src/wam2layers/tracking/backtrack.py
ruudvdent: done

Peter9192 (approved):
Nice work. Can be polished a bit in the future.

PR #319, opened 2024-02-05
Add plots of the difference in the regression tests

BSchilperoort:
To make it easier to quickly see what went wrong, I expanded the workflow tests with a plotting function.
Plots of the difference between the expected and actual outputs are generated for the preprocessing and backward tracking tests.

I also fixed a small mistake in analysis/checks.py: The stackoverflow answer was (slightly) incorrect, which was detected by [mypy](https://mypy-lang.org/).

### DIFF
--- a/tests/test_workflow.py
+++ b/tests/test_workflow.py
@@ -1,12 +1,51 @@
 """Tests for entire workflows."""
 from pathlib import Path
 
+import matplotlib.colors
 import matplotlib.testing.compare
 import numpy
 import xarray as xr
 from click.testing import CliRunner
+from matplotlib import pyplot as plt
 
 from wam2layers.cli import cli
+
+try:  # Use cartopy if it's available
+    import cartopy.feature as cfeature
+    from cartopy import crs
+except ImportError:
+    crs = None
+    cfeature = None
+
+
+def plot_difference(
+    expected_output: xr.DataArray, output: xr.DataArray, fpath: Path
+) -> None:
+    """Plot the difference between the expected output and actual output."""
+    if not fpath.parent.exists():
+        fpath.parent.mkdir()  # Make figures directory if it does not exist yet
+
+    if crs is not None:
+        subplot_kw = dict(projection=crs.PlateCarree())
+    else:
+        subplot_kw = None
+
+    fig, ax = plt.subplots(
+        figsize=(16, 10),
+        subplot_kw=subplot_kw,
+    )
+    (expected_output - output).plot.pcolormesh(
+        ax=ax,
+        x="longitude",
+        cmap="RdBu",
+        norm=matplotlib.colors.TwoSlopeNorm(vcenter=0),
+    )
+    if cfeature is not None:
+        ax.add_feature(cfeature.COASTLINE, linewidth=0.8)
+        ax.add_feature(cfeature.BORDERS, linestyle="-", linewidth=0.2)
+
+    ax.set_title(f"Difference between expected and actual '{output.name}' values.")
+    fig.savefig(fpath, dpi=200)
 
 
 def test_preprocess():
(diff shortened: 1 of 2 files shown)


tests/test_workflow.py
```diff
@@ -1,13 +1,52 @@
...
+
+    if crs is not None:
+        subplot_kw = dict(projection=crs.PlateCarree())
+    else:
+        subplot_kw = None
+
+    fig, ax = plt.subplots(
+        figsize=(16, 10),
+        subplot_kw=subplot_kw,
+    )
```
Peter9192, line 36: Although I like your solution for "maybe import cartopy", this gives me typing issues from pylance/pyright: 
```
"Argument of type "dict[str, PlateCarree] | None" cannot be assigned to parameter "subplot_kw" of type "dict[Unknown, Unknown]" in function "subplots""
```
You could set `subplot_kw` to `{}` instead of `None` and adapt the check on line 43 accordingly.

tests/test_workflow.py
```diff
@@ -1,13 +1,52 @@
...
+    (expected_output - output).plot.pcolormesh(
```
Peter9192, line 37: This also triggers pylance for me, though I'm not really sure why: `Argument missing for parameter "self"`

tests/test_workflow.py
```diff
@@ -22,11 +61,20 @@ def test_preprocess():
...
+        output["precip"].sum("time"),
+        diff_figure_path,
+    )
+
     numpy.testing.assert_allclose(
         expected_output["precip"].values,
         output["precip"].values,
         err_msg=(
             "Output results are different! Please verify your results. \n"
+            f"A difference plot is available under {diff_figure_path}. \n"
```
Peter9192, line 77: Just a note, but further down below, on line 88, I get a failing check if there were log files present from previous test runs. Perhaps we should clean them up at the end of the tests? We could address that here or in another PR

Peter9192 (approved):
Nice work! I get a few red wiggles from pylance, perhaps it's my environment. Also,

* could you add an entry to the changelog?
* do you think we could upload the figures as artificats in the GH actions runner, so it is even easier to inspect them from a PR?

Feel free to merge after another pass

tests/test_workflow.py
BSchilperoort: AFAIK I did not introduce the failing check if log files were previously present. That was already there. In the next PR I intend to improve how the log files are checked.

tests/test_workflow.py
BSchilperoort: Strange, the default value for `subplot_kw` is `None`. https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.subplots.html

Perhaps this is incorrectly typed in Microsoft's unofficial matplotlib stubs. Mypy doesn't detect a problem here.

tests/test_workflow.py
BSchilperoort: No idea what is going wrong here. That again seems like a bugged type in pylance. Mypy does not detect this.

BSchilperoort:
> * could you add an entry to the changelog?
> 
>     * do you think we could upload the figures as artificats in the GH actions runner, so it is even easier to inspect them from a PR?

Done and done.

The artifacts are available when you look at the tests. You can download the (zipped) plots there. E.g.:

```
Artifact diff-plots_py3.8-ubuntu-latest.zip successfully finalized. Artifact ID 1223262701
Artifact diff-plots_py3.8-ubuntu-latest has been successfully uploaded! Final size is 595182 bytes. Artifact ID is 1223262701
Artifact download URL: https://github.com/WAM2layers/WAM2layers/actions/runs/7798384693/artifacts/1223262701
```

Issue #317, opened 2024-01-24
coordinate bug in correction debugger


ruudvdent:
I was running a forward tracking test experiment. I got an error the moment the stability correction is invoked. Something with writing the correction file to netcdf doesn't properly with the coordinates. In fact the longitude has length 281 and latitude 161.

````
wam2layers forwardtrack  Moisture_tracking_cases/West_Africa_1998_JA.yaml 
Welcome to WAM2layers.
Starting forwardtrack experiment.
Initializing experiment with config file Moisture_tracking_cases/West_Africa_1998_JA.yaml
Output will be written to /home/rjvanderent/Moisture_tracking_cases/West_Africa_JA/output_data.
1998-07-02 00:00:00 - Tracked moisture: 1.85%. Tagged in atmosphere: 98.15%. Lost moisture: 0.00%. Gained moisture: 0.00%. Tagging closure: 100.00%. Time since start: 33.09s, RAM: 49.61 MB
1998-07-02 00:00:00 - Writing output to file /home/rjvanderent/Moisture_tracking_cases/West_Africa_JA/output_data/forwardtrack_1998-07-02T00-00.nc
1998-07-03 00:00:00 - Tracked moisture: 9.99%. Tagged in atmosphere: 90.01%. Lost moisture: 0.00%. Gained moisture: 0.00%. Tagging closure: 100.00%. Time since start: 71.11s, RAM: 63.06 MB
1998-07-03 00:00:00 - Writing output to file /home/rjvanderent/Moisture_tracking_cases/West_Africa_JA/output_data/forwardtrack_1998-07-03T00-00.nc
Traceback (most recent call last):
  File "/home/rjvanderent/miniconda3/envs/wam_env/bin/wam2layers", line 8, in <module>
    sys.exit(cli())
             ^^^^^
  File "/home/rjvanderent/miniconda3/envs/wam_env/lib/python3.11/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/rjvanderent/miniconda3/envs/wam_env/lib/python3.11/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/home/rjvanderent/miniconda3/envs/wam_env/lib/python3.11/site-packages/click/core.py", line 1688, in invoke
[...]

PR #325, opened 2024-02-14
Fix bug in profiling

### LINKED ISSUE #324: Bug in profiling's `track_stability_correction`
I tried to run the lake Volta test case, but encountered a bug which prevented the code from continuing past day 8.

It turns out that the [code here](https://github.com/WAM2layers/WAM2layers/blob/a53f4c46a79f990feaa56a10bd3b6ebd6040866e/src/wam2layers/utils/profiling.py#L132) assumes that the `coords` are in the same order as the dimensions of `correction`.
However, in my case they are not. Transposing the `correction` matrix makes the code run OK.

```py
ncfile = xr.DataArray(
    correction, # `correction.T` for me.
    coords=coords, 
    name="correction",
)
```
However, a more robust fix is probably needed.

Seems like the following should work. Replace:
```py
        correction = np.where(corrected, fy_abs - fy_corrected, 0)
```
With:
```py
        correction = xr.where(corrected, fy_abs - fy_corrected, 0)
```
Which returns a DataArray, so coordinates are kept.

BSchilperoort:
Fixes #324

### DIFF
--- a/src/wam2layers/utils/profiling.py
+++ b/src/wam2layers/utils/profiling.py
@@ -1,12 +1,11 @@
 import logging
 import time
-from datetime import datetime
 
-import numpy as np
+import pandas as pd
 import psutil
 import xarray as xr
 
-from wam2layers.tracking.io import load_tagging_region
+from wam2layers.config import Config
 
 logger = logging.getLogger(__name__)
 
@@ -97,7 +96,14 @@ class ProgressTracker:
             f"Time since start: {time}s, RAM: {memory:.2f} MB"
         )
 
-    def track_stability_correction(self, fy_corrected, fy_abs, config, coords, t):
+    def track_stability_correction(
+        self,
+        fy_corrected: xr.DataArray,
+        fy_abs: xr.DataArray,
+        config: Config,
+        coords: xr.Coordinates,
+        t: pd.Timestamp,
+    ) -> None:
         """Issue warning if correction exceeds criterion.
 
         Warning advises to reduce the timestep.
@@ -109,7 +115,7 @@ class ProgressTracker:
         """
         corrected = fy_corrected < fy_abs
         corrected_percent = corrected.sum() / corrected.count() * 100
-        correction = np.where(corrected, fy_abs - fy_corrected, 0)
+        correction = xr.where(corrected, fy_abs - fy_corrected, 0)
         correction_max = correction.max()
 
         # Reversed conditions lead to cleaner code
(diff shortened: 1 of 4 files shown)


src/wam2layers/utils/profiling.py
```diff
@@ -129,12 +135,12 @@ def track_stability_correction(self, fy_corrected, fy_abs, config, coords, t):
...
-        ncfile = xr.DataArray(correction, coords=coords, name="correction")
-        ncfile.to_netcdf(filename)
+        correction.name = "correction"
+        correction.to_netcdf(filename)
 
         logger.warn(
             f"Stability correction applied to {corrected_percent:.1f}% of "
             f"grid. Average correction was {correction.mean():.1f}, "
-            f"maximum correction was {correction.max()}. The total "
+            f"maximum correction was {correction.max():.1f}. The total "
```
Peter9192, line 144: Not sure if we should limit the absolute value to 1 decimal place, perhaps use scientific notation?

src/wam2layers/utils/profiling.py
```diff
@@ -97,7 +96,14 @@ def print_progress(self, t, output):
             f"Time since start: {time}s, RAM: {memory:.2f} MB"
         )
 
-    def track_stability_correction(self, fy_corrected, fy_abs, config, coords, t):
+    def track_stability_correction(
+        self,
+        fy_corrected: xr.DataArray,
+        fy_abs: xr.DataArray,
+        config: Config,
+        coords: xr.Coordinates,
```
Peter9192, line 104: The python3.8 tests don't like this type annotation. Should we stop supporting it?

Peter9192 (approved):
Thanks for the fix. I think it also closes #317.

src/wam2layers/utils/profiling.py
BSchilperoort: I guess we can do that. I mostly copied the formatting from the previous lines so it would format it as a float and not display the xarray object.

src/wam2layers/utils/profiling.py
BSchilperoort: The type notation fails because `xr.Coordinates` was added 7 months ago, and xarray stopped supporting 3.8 a bit over a year ago. Xarray follows numpy on this, so py3.8 will use old numpy versions as well.

I think following their version restrictions would be best (i.e. deprecate support for python versions when xarray/numpy drops it)

src/wam2layers/utils/profiling.py
Peter9192: Then we could also use `.item()` or `.values()`

src/wam2layers/utils/profiling.py
Peter9192: > I think following their version restrictions would be best (i.e. deprecate support for python versions when xarray/numpy drops it)

Sounds reasonable to me



> We probably also should specify `xarray>=2023.08.0`, that's when the coordinate object was added to the public API

I would not be very happy with that. I imagine many researchers may want to install this in an existing environment, and requiring such a recent release might lead to unnecessary conflicts. The types are not used at runtime anyway.

BSchilperoort: > The types are not used at runtime anyway

OK but then I will have to do:
```py
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from xarray import Coordinates

def func(
        coords: "Coordinates",
):
```

otherwise it will still mess with code at runtime

BSchilperoort: I just realized that the coords are not used anymore so there's no need for the Coordinate object.

Issue #280, opened 2023-11-08
units and details in onscreen logging output


ruudvdent:
We might get:

Stability correction applied to 16.5% of grid. Average correction was 45532.9, maximum correction was 1689801.5562748688. The total correction field is written to /home/rjvanderent/Moisture_tracking_cases/floodcase_202107_new_fv/output_data/debug/stability_correction_20210715-000000.nc.
2021-07-15 00:00:00 - Tracked moisture: 0.01%. Lost moisture: -0.01%. Time since start: 0.79s, RAM: 20.82 MB
2021-07-15 00:00:00 - Writing output to file /home/rjvanderent/Moisture_tracking_cases/floodcase_202107_new_fv/output_data/backtrack_2021-07-15T00-00.nc

Add info:

- unit of the correction
- when new stability correction info will be given
- tracked moisture in %, provide info on percentage of what
- lost moisture in %, provide info on percentage of what 
- lost moisture field (can it be found somewhere, or what logging setting should be set in the yaml to store it?)

Peter9192:
The logging is now still a bit ad-hoc. It would be nice if the logger had access to the full state of the model. That would make it much easier to change things like this. We could consider a more object-oriented approached. Also touches on #260

ruudvdent:
Regarding losses this is solved by #305 , but not yet for the stability correction

PR #332, opened 2024-02-19
Modify stability correction warning

BSchilperoort:
When working together with Ruud, we noticed some issues with the stability correction warning.

The checks have been modified slightly, and will now alert the user when:

1. At least 5% of the grid was corrected (first time), or the area corrected has increased by 10% (afterwards).
2. The correction factor (its value is 0 - 1, lower numbers is stronger correction) is 10% lower (i.e. stronger).

### DIFF
--- a/src/wam2layers/utils/profiling.py
+++ b/src/wam2layers/utils/profiling.py
@@ -37,8 +37,8 @@ class ProgressTracker:
         self.gained_water = 0
         self.store_intermediate_states(output)
         self.profile = Profiler()
-        self.stability_correction_previous_grid = 0
-        self.stability_correction_previous_value = 0
+        self.stability_correction_previous_grid = 5/1.1
+        self.stability_correction_previous_value = 1
 
     def store_intermediate_states(self, output):
         """Update moisture totals based on output since previous time step.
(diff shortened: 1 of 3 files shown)


src/wam2layers/utils/profiling.py
```diff
@@ -109,37 +109,44 @@ def track_stability_correction(
 
         Criteria:
         1. A correction is applied for more than 5% of the grid (first time)
-        2. Previous correction is exceeded by either 5%-point of grid
-        3. Previous correction is doubled in magnitude
+        2. At least 10% more of the grid is corrected compared to the last warning
```
Peter9192, line 112: Isn't that quite often? So if I first get a warning at 5%, I will get the next one at 5.5%, and then at 6.05%, etc. I would make it at least 50% if relative, or stick to using increases in %-points

src/wam2layers/utils/profiling.py
```diff
@@ -109,37 +109,44 @@ def track_stability_correction(
...
+        3. Previous strongest correction factor is exceeded by 10%
```
Peter9192, line 113: Here as well. I would expect this leads to much more warnings. Is that what you wanted?

Peter9192 (approved):
> we noticed some issues with the stability correction warning

Can you explain a bit more what "the issues" were? 

If I understand it correctly, this change will mostly lead to much more warnings for much smaller increments in the applied correction. Was that the intention?

The improved warning seems good to me.

src/wam2layers/utils/profiling.py
ruudvdent: A few more warnings and 2D-grids to analyze are not so problematic, but rather informative, I suppose. The main issue was, however, to provide more informative on-screen logging. See #280

ruudvdent:
closes #280

ruudvdent (approved):
Excellent, this new onscreen info is much more intuitive to understand for the user.

BSchilperoort:
@Peter9192 it will lead to some more warnings, especially at the start of a run (at least in the west-africa forward tracking case).
If it turns out the warnings occur too often now, we can adjust the factors. For example a 20% increase instead of the current 10% increases

> Can you explain a bit more what "the issues" were?

- The input for the function was `fy_limit` instead of `fy_stable` (the function wanted `fy_corrected` which is the second one
- We now compute the correction factor instead of the absolute correction value (which was difficult to interpret)
- The two trigger conditions overwrote each other's stored value
  - For example, the percentage of grid exceeds 5%, but the correction was quite minor. This resets the "correction_max" to a lower value, which could be easily triggered a different time step. Which would in turn set the stored `corrected_percent` to a lower value.

Peter9192:
Great, that makes a lot of sense. Thanks for the update!

PR #334, opened 2024-02-21
Improve output files

### LINKED ISSUE #287: More descriptive output files
It would be great if we could clearly communicate the meaning of the output variables, including their units ("accumulation since previous time step" etc), by using the units and attrs fields in the netcdf file.

BSchilperoort:
Closes #287 

This PR adds dataset attributes to the output files, as well as a time dimension. This allows for easier to interpret files and easier loading.

You can now open all output files for analysis using:
```py
import xarray as xr
ds = xr.open_mfdataset("path_to_output_folder/backtrack*.nc")
```
The `ds` object will have a lot of info added to the dataset itself and the variables. With the HTML display in a jupyter notebook these attributes are easy to view:

<details><summary> preview </summary>

image

</details>


src/wam2layers/reference/variables.py
ruudvdent, line 1: This is a really useful addition. I was wondering about only 1 thing. Suppose you have only 1 output file, there is no obvious way to infer the 'output time step'. Hence, if one has accumulated tracked evaporation, one would have to place an educated guess over what time it was accumulated. Doesn't have to be in this PR, but would be nice to make to somehow even more clear.

ruudvdent (approved):

src/wam2layers/reference/variables.py
BSchilperoort: Good point. In the CF conventions, this is usually done by adding "bounds":

> It is often the case that data values are not representative of single points in time and/or space, but rather of intervals or multidimensional cells. This convention defines a bounds attribute to specify the extent of intervals or cells. 

See http://cfconventions.org/cf-conventions/v1.6.0/cf-conventions.html#cell-boundaries

This would look like:
image

But this can be added later, as long as the actual time coordinates do not change.

BSchilperoort: For now I have added the (serialized) config file to the attributes. This contains the output time step as well.

PR #338, opened 2024-03-11
Change CLI to `wam2layers track configfile.yml`

### LINKED ISSUE #328: Finalize cli visualize commands, support forwardtracking
Before v3 the interface for visualization should be finalized.

A test for the visualization of forward tracking data should also be added.

BSchilperoort:
Instead of trying to infer the tracking direction in, e.g., the visualization code, it's easier to put the direction in the config file.
Now that's changed, we don't need to specify "forwardtrack" or "backtrack" anymore in the CLI. Closes #328


Peter9192:
I'm all for this. Just wonder if 'tracking direction' is the best name for the new config setting. I was thinking of 'tracking mode' as an alternative. 

@ruudvdent are time and distance tracking alternative modes or additional options for a forward or backward run?

ruudvdent:
I think tracking direction is clear, mode would be fine as well, so no particular preference from my side. Time and distance tracking would rely on the basic tracking in either direction for the calculations, so they would be additional options.

Peter9192:
ok let's stick with direction then. Can this be merged?

Peter9192 (approved):
I tried running the example floodcase again, and this works nicely. Note that we need to update the example config [here](https://rebrand.ly/wam2layers-example-data) as well.

Peter9192:
> I tried running the example floodcase again, and this works nicely. Note that we need to update the example config [here](https://rebrand.ly/wam2layers-example-data) as well.

I have updated this file

PR #342, opened 2024-03-27
Add publishing action, update dev docs

### LINKED ISSUE #180: Automate publish on PyPI
It is possible to automatically build and push the package to PyPI on every release. To this end, you can add a [GitHub action](https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/) that is triggered on a `release` event. See this PR for an example: https://github.com/eWaterCycle/ewatercycle/pull/327

BSchilperoort:
Closes #180 

- I followed the instructions on https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/
- On PyPI I made the workflow file a "Trusted Publisher".
- I copied the workflow from https://github.com/eWaterCycle/ewatercycle-leakybucket where it ran successfully yesterday.

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -1,0 +1,49 @@
+name: Publish Python 🐍 distribution 📦 to PyPI
+
+on:
+  release:
+    types: [created]
+
+jobs:
+  build:
+    name: Build distribution 📦
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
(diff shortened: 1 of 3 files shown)


Peter9192 (approved):
Very cool! We'll only know for sure that it works once we make a new release, but from the looks of it I think everything should be okay. I also verified that I can see the trusted publisher on PyPI.

Can you add a changelog entry?

PR #343, opened 2024-03-27
Split logged losses in boundary-transport and interior losses

### LINKED ISSUE #340: Logged "losses" now comprises of "boundary losses" and "unaccounted water"
I just ran the example floodcase again, and noticed the new logging behaviour from #305. I was surprised and worried to see `Lost moisture: 52.00%.`

Upon inspection of the code, I realized that the "lost moisture" now contains both "boundary losses" and "unaccounted water". I think this is slightly confusing: boundary losses are expected and nothing to worry about, even if it's a really high value. Whereas large amounts of unaccounted water is definitely a reason for concern.

Should we split them again?

Peter9192:
Closes #340 

Now the (last) output of the sample run is:
```
2021-07-01 00:00:00 - Tracked moisture: 44.38%. Tagged in atmosphere: 3.63%. Boundary transport: 52.00%. Lost moisture: 0.00%. Gained moisture: 0.00%. Tagging closure: 100.00%. Time since start: 248.96s, RAM: 63.20 MB
2021-07-01 00:00:00 - Writing output to file /home/peter/wam2layers/floodcase_202107/output_data/backtrack_2021-07-01T00-00.nc
Experiment complete.
```

### DIFF
--- a/src/wam2layers/utils/profiling.py
+++ b/src/wam2layers/utils/profiling.py
@@ -35,6 +35,7 @@ class ProgressTracker:
         self.tracked = 0
         self.lost_water = 0
         self.gained_water = 0
+        self.boundary_transport = 0
         self.store_intermediate_states(output)
         self.profile = Profiler()
         self.stability_correction_previous_grid = 0.0
@@ -47,14 +48,21 @@ class ProgressTracker:
         don't lose accumulations from previous outputs.
         """
         totals = output.sum()
+
         if self.mode == "backtrack":
             self.tracked += totals["e_track"]
             self.total_tagged_moisture += totals["tagged_precip"]
         else:  # mode is forwardtrack
             self.tracked += totals["p_track_lower"] + totals["p_track_upper"]
             self.total_tagged_moisture += totals["tagged_evap"]
+
+        # Don't include boundary losses in log messages
+        interior_losses = output.losses[1:-1, 1:-1].sum()
+        boundary_transport = totals["losses"] - interior_losses
+
+        self.lost_water += interior_losses
+        self.boundary_transport += boundary_transport
         self.gained_water += totals["gains"]
-        self.lost_water += totals["losses"]
 
     def print_progress(self, t, output):
         """Print some useful runtime diagnostics."""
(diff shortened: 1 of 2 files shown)


src/wam2layers/utils/profiling.py
```diff
@@ -47,14 +48,21 @@ def store_intermediate_states(self, output):
...
+
         if self.mode == "backtrack":
             self.tracked += totals["e_track"]
             self.total_tagged_moisture += totals["tagged_precip"]
         else:  # mode is forwardtrack
             self.tracked += totals["p_track_lower"] + totals["p_track_upper"]
             self.total_tagged_moisture += totals["tagged_evap"]
+
+        # Don't include boundary losses in log messages
+        interior_losses = output.losses[1:-1, 1:-1].sum()
```
BSchilperoort, line 60: As you use a simple mask to split interior/boundary, I guess this means that you can theoretically overestimate "boundary losses" and underestimate interior losses :thinking: 
But that's probably more of an academic point and doesn't really matter in practice.

BSchilperoort (approved):
Nice simple fix for this issue! :+1:

src/wam2layers/utils/profiling.py
ruudvdent: I rather include both in the log messages, so that a user can see what's happening.

ruudvdent (approved):

BSchilperoort:
In the output file there is the variable "Lost moisture". This is including boundary-transport, right?

So perhaps it's better to clarify that the `Lost moisture:` in the log file represents only internal losses.

Peter9192:
> As you use a simple mask to split interior/boundary, I guess this means that you can theoretically overestimate "boundary losses" and underestimate interior losses 🤔
> But that's probably more of an academic point and doesn't really matter in practice.

Indeed, the decision to combine them was made in #305.

> In the output file there is the variable "Lost moisture". This is including boundary-transport, right?

yes

> So perhaps it's better to clarify that the `Lost moisture:` in the log file represents only internal losses.
 
I considered logging "boundary losses" and "interior losses", but decided against it to keep the contrast between losses and gains.

We can reconsider all these points, but I will merge this for now as I think it is already a good improvement and it fixes #340

PR #346, opened 2024-03-27
Rename `target_frequency` to `timestep` and move it to core computations

Peter9192:
This PR renames the configuration parameter `target_frequency` to `timestep`. This name is more recognizable as it is one of the key variables in any model. 

Additionally, the location where this parameter is used is changed. Before, we used to apply it once in the `change_units` function, such that everywhere else in the code, the fluxes represented accumulated quantities in units of m3. Now,  the timestep is applied in the core computation routines, i.e. $\partial S / \partial t = F$ is implemented as $S^{t+1} = S^t + \Delta t * F$. In my opinion, this makes it much easier to see how the formula on paper is related to the implementation in the code.


BSchilperoort (approved):
Nice work, it all looks good to me!

Issue #323, opened 2024-02-14
Preprocessing vs tracking


Peter9192:
Currently we are doing some things in the tracking module that would fit better with the preprocessing. For example, changing the units. Most importantly, I think it makes sense for all fluxes, including precip and evap, to be fed into the tracking code as "rates", i.e. units m3/s, instead of accumulations. Then,

* The tracking code will be independent of the frequency of the input data
* The time step used in tracking can be incorporated in the numerical integration step

Since this affects the backward compatibility of preprocessed data created with older versions of WAM2layers, ideally this should be decided upon and implemented before the V3 release.

ruudvdent:
After discussion with @Peter9192 and @BSchilperoort 

- preprocessing is already in rates and standard units like [kg m-2 s-1] for E and P, and [kg m-1 s-1] for F 
- next step would be to investigate if we can drop to conversion to m3 in the tracking completely. Tracking in mass is still preferred from a point of view of following with happens, e.g., loss calculations, which would be cumbersome when tracking concentrations. Major consequence would be that the accumulated output E_track and P_track would also be in other units: [kg m-2}.

PR #349, opened 2024-03-27
Calculate with kgs instead of m3

Peter9192:
This PR changes the internal calculation units to kgs. The division by density of liquid water was not necessary and unnecessarily complicates the code as well as the description in the paper. Also, m3s are hard to interpret. Kgs are easier to convert back to kg/m3 or mm.


Peter9192:
Test difference is exactly a factor 1000:

image
image

Peter9192:
From https://github.com/WAM2layers/WAM2layers/issues/323#issuecomment-1948147556
> next step would be to investigate if we can drop to conversion to m3 in the tracking completely. Tracking in mass is still preferred from a point of view of following with happens, e.g., loss calculations, which would be cumbersome when tracking concentrations. Major consequence would be that the accumulated output E_track and P_track would also be in other units: [kg m-2}.

So far I have only dropped the division by density. We still multiply by area though in the `change_units` function. That means that we are now calculating in [kg]s. Alternatively, we could calculate in [kg m-2].

Advantages: 

* The output values would be in units of [kg m-2] as well, as suggested by Ruud in the quote above. This is more interpretable than bare kgs
* It is consistent with other literature; e.g. [trenberth 1995](https://doi.org/10.1175/1520-0442(1995)008%3C2255:EOTGAM%3E2.0.CO;2) who define the "precipitable water" as the "vertical mass-weighted integral specific humidity", $w=\int q\rho dz = q\rho\Delta z = q\Delta p / g$, i.e. exactly what we are calculating in the pre-processing.
* We could skip the multiplication by area for precip, evap, s_upper and s_lower. 
* The multiplication by dx, dy in `change_units` would change to a division by the other. In maths, $F_{x,new} = F_{x,old}/A = q\rho\Delta y\Delta z/A = q\rho\Delta z/\Delta x = w/\Delta x$. In this way, the advection terms are more recognizable: $\nabla \cdot w = \partial w/\partial x + \partial w/\partial y$


Potential disadvantage:

*  In checking the closure of the moisture budget, we still need to multiply by the area, as the source and sink regions are associated with different grid cells, and thus different areas.

Peter9192:
Output is now in [kg m-2], i.e. roughly a factor 1e8 different wrt the [kg] units (grid cells of roughly 30 km * 30 km), or 1e5 different from the old units of [m3]. This can be seen from the test output:

```
Mismatched elements: 1301 / 1681 (77.4%)
Max absolute difference: 7511.107
Max relative difference: 700307.
```

Interestingly, after removing the conversion to mm from the visualization routines, the visualization tests pass again, so the output figures are exactly the same as before. 

Calculating in kg/m2 instead of kg introduces a difference in the vertical flux, as can be seen from the difference plot:

backtrack_diff

It looks like the convergence that's calculated as part of the vertical flux produces slightly different results. Possibly, this is due to slightly different Fx and Fy inputs caused by the stability correction. 

Due to this, the water balance no longer closes for the 2021 flood case. I now get 

```
2021-07-12 00:00:00 - Tracked moisture: 16.23%. Tagged in atmosphere: 92.21%. Boundary transport: 0.00%. Lost moisture: 0.00%. Gained moisture: 0.00%. Tagging closure: 108.44%. Time since start: 52.06s, RAM: 29.25 MB
```

Still looking into that.

ruudvdent:
> ```
> 2021-07-12 00:00:00 - Tracked moisture: 16.23%. Tagged in atmosphere: 92.21%. Boundary transport: 0.00%. Lost moisture: 0.00%. Gained moisture: 0.00%. Tagging closure: 108.44%. Time since start: 52.06s, RAM: 29.25 MB
> ```

That's weird, but I am not sure how this works now. The closure should be in absolute mass then, thus kg and not per unit area obviously. I think more than just Fx and Fy being slightly different is going on.

Peter9192:
The systematic pattern was already present upon the removal of density. So I think it's an artefact of the visualization, and not a systematic difference after all.

The closure of the moisture balance was due to an error on my sight. I corrected for the area in the balance calculations, but this is not necessary. My bad.

Peter9192:
Reverting the last commit gives a closed moisture balance. All problems solved.

I will close this in favour of a clean PR (#356)

PR #351, opened 2024-04-02
Move CI away from Conda

### LINKED ISSUE #348: Move CI to not use conda
Cartopy now has wheels available on PyPI (meaning that you can just use pip to install it, no conda needed, see #345).

This means that we also don't need micromamba on the CI, and can instead use pip.

We could also decide to use [uv](https://github.com/astral-sh/uv) to speed up the CI even more. This would just mean replacing any `pip` command with `uv pip`. 
For example, `uv pip install wam2layers[dev]`.

A complete install with `pip install uv; uv pip install wam2layers[complete]` took 36 s on my system (clean venv).
After removing that venv and doing `pip install wam2layers[complete]` in a new venv, it took 54 s.

BSchilperoort:
closes #348

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -21,26 +21,28 @@ jobs:
     env:
       MPLBACKEND: Agg  # https://github.com/orgs/community/discussions/26434
     steps:
-      - uses: actions/checkout@v3
-      - uses: mamba-org/setup-micromamba@v1
+      - uses: actions/checkout@v4
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v4
         with:
-          environment-name: wamenv
-          create-args: >-
-            python=${{ matrix.python-version }}
-            cartopy
-          init-shell: bash
-      - name: Micromamba info
-        shell: bash -el {0}
+          python-version: ${{ matrix.python-version }}
+      - name: Python info
+        shell: bash -l {0}
         run: |
-          micromamba info
-      - name: Install dev dependencies
+          which python3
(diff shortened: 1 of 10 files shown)


.github/workflows/ci.yaml
```diff
@@ -21,26 +21,25 @@ jobs:
...
-          wam2layers --help
-        shell: bash -el {0}
-      - name: Run pytest
-        run: pytest --basetemp=tmp/
-        shell: micromamba-shell {0}
+          python3 -m pip install --upgrade pip
+      - name: Install WAM2Layers
+        run: |
+          python3 -m pip install .[dev]
+      - name: Run unit tests
```
Peter9192, line 40:
```suggestion
      - name: Run tests
```

.github/workflows/ci.yaml
```diff
@@ -11,7 +11,7 @@ on:
 jobs:
 
   build:
-    name: Micromamba test on (${{ matrix.python-version }}, ${{ matrix.os }})
+    name: Unit tests on (${{ matrix.python-version }}, ${{ matrix.os }})
```
Peter9192, line 14:
```suggestion
    name: Run tests on (${{ matrix.python-version }}, ${{ matrix.os }})
```

Peter9192 (approved):
Nice to get rid of mamba. Simpler = better. Most important tests are regression tests rather than unit tests :-)

Peter9192:
Shall we add a changelog entry for this as well?

BSchilperoort:
Thanks for the review, @Peter9192 !

PR #352, opened 2024-04-02
Add --version option

BSchilperoort:
Modeled after pip: https://github.com/pypa/pip/blob/02ad0f3ffd278aa0a856cb9d80e0425d36838d1a/src/pip/_internal/utils/misc.py#L76

To test, do:
`wam2layers --version`

This should display the version, along with where it was installed and the python version (useful for debugging).

We could consider adding this to the issue template


Peter9192 (approved):
Very nice! Can you add a changelog entry?

Issue #347, opened 2024-03-27
Reference time inconsistent between forward and backward tracking


Peter9192:
In backtrack, the variable `S0` is unused, whereas in forwardtrack, both `S1` and `S0` are used. This seems inconsistent.

https://github.com/WAM2layers/WAM2layers/pull/346/commits/30fc08853cb8faa477cf8d33a0f495e81b5410b3

Peter9192:
Closed in #355

PR #355, opened 2024-04-04
Bookkeeping

Peter9192:
This PR isolates some commits from #349 concerning the bookkeeping of the tracking closure. Refs #347

* Reference state for backtracked is changed from `S1` to `S0`, i.e. the total moisture in the grid cell at the end of the timestep
* Gains are calculated as absolute fields instead of relative to the total moisture
* Gains are set to be positive (up for debate, but should be consistent with budget calcuation in log message)
* Warning about gains is moved to a separate utility function such that the tracking code is easier to read
* Bookkeeping and output writing are separated more clearly in the code.

Open questions (@ruudvdent): 
* Why not calculate the gains *after* redistributing unaccounted moisture between upper and lower?

### DIFF
--- a/src/wam2layers/tracking/backtrack.py
+++ b/src/wam2layers/tracking/backtrack.py
@@ -25,7 +25,7 @@ from wam2layers.tracking.io import (
     write_output,
 )
 from wam2layers.tracking.shared import initialize_tagging_region, initialize_time
-from wam2layers.utils.profiling import ProgressTracker, check_for_gains
+from wam2layers.utils.profiling import ProgressTracker, warn_about_gains
 
 logger = logging.getLogger(__name__)
 
@@ -79,10 +79,13 @@ def backtrack(
     )
 
     # account for negative storages that are set to zero: "numerically gained water"
-    # TODO: reference should be S0?
-    s_track_lower, gains_lower = check_for_gains(s_track_lower, reference=S0["s_lower"])
-    s_track_upper, gains_upper = check_for_gains(s_track_upper, reference=S0["s_upper"])
+    # TODO: why divide by S1/S0? I think it should be absolute values
+    gains_lower = np.where(s_track_lower < 0, s_track_lower / S0["s_lower"], 0)
+    gains_upper = np.where(s_track_upper < 0, s_track_upper / S0["s_upper"], 0)
     gains = np.abs(gains_lower + gains_upper)
+    s_track_lower = np.maximum(s_track_lower, 0)
+    s_track_upper = np.maximum(s_track_upper, 0)
+    warn_about_gains(gains_lower, gains_upper)
 
     # lower and upper: redistribute unaccounted water that is otherwise lost from the sytem
     overshoot_lower = np.maximum(0, s_track_lower - S0["s_lower"])
--- a/src/wam2layers/tracking/backtrack.py
+++ b/src/wam2layers/tracking/backtrack.py
@@ -26,7 +26,7 @@ from wam2layers.tracking.io import (
     write_output,
 )
 from wam2layers.tracking.shared import initialize_tagging_region, initialize_time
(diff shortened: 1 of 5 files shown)


Peter9192:
@ruudvdent would you be able to have a look at this? I'm not entirely certain whether some of the old code was intentional or accidental, so feel free to correct me if I misinterpreted things.

src/wam2layers/tracking/backtrack.py
```diff
@@ -25,15 +26,15 @@
     write_output,
 )
 from wam2layers.tracking.shared import initialize_tagging_region, initialize_time
-from wam2layers.utils.profiling import ProgressTracker
+from wam2layers.utils.profiling import ProgressTracker, warn_about_gains
 
 logger = logging.getLogger(__name__)
 
 
 def backtrack(
```
BSchilperoort, line 34: A large amount of code is shared between the forwardtrack and backwardtrack functions now (13 lines at the start of the functions, gains accounting, output updating...). Seeings as these lines are identical, could this be refactored in a way to reduce the duplication? 

The initialize functions in each module are also nearly identical.

`run_experiment` is also nearly duplicated:

<details><summary>Comparing run_experiment</summary>
<p>

image

</p>
</details> 

I guess it's a balancing act between readability and reducing duplication, so I was wondering what your view on this is.

src/wam2layers/tracking/backtrack.py
```diff
@@ -85,54 +78,41 @@ def backtrack(
...
-    s_track_upper = np.maximum(s_track_upper, 0)
 
     # account for negative storages that are set to zero: "numerically gained water"
-    gains = np.abs(s_track_negative_lower + s_track_negative_upper)
+    gains_lower = np.where(s_track_lower < 0, -s_track_lower, 0)
+    gains_upper = np.where(s_track_upper < 0, -s_track_upper, 0)
+    gains = np.abs(gains_lower + gains_upper)
+    s_track_lower = np.maximum(s_track_lower, 0)
+    s_track_upper = np.maximum(s_track_upper, 0)
+    warn_about_gains(gains_lower, gains_upper)
```
Peter9192, line 88: Let's remove the warning, as it is already logged during output write

Peter9192:
in profiling, gains should be added

Peter9192:
> * Why not calculate the gains _after_ redistributing unaccounted moisture between upper and lower?

We decided to shift it

src/wam2layers/tracking/backtrack.py
```diff
@@ -25,15 +26,15 @@
     write_output,
 )
 from wam2layers.tracking.shared import initialize_tagging_region, initialize_time
-from wam2layers.utils.profiling import ProgressTracker
+from wam2layers.utils.profiling import ProgressTracker, warn_about_gains
```
ruudvdent, line 30: I think we've discussed to remove this warn_about_gains

src/wam2layers/tracking/backtrack.py
ruudvdent: I suggest we make an issue "reduce duplication", but leave it for now

ruudvdent (changes requested):
cool stuff, just implementing the points as discussed and we're good to go!

src/wam2layers/tracking/backtrack.py
Peter9192: I've been working towards this. The fact that they are now already almost the same means we've been quite succesful in aligning them. I agree working towards more dupe reduction is a good goal. Ideally I want the main "run experiment" functions to be something like 

```py
from wam2layers import model

m = model()
m.initialize(config_file)
while model < model.start_time # this is different for forward/backward
    model.update
    if ... write output
model.finalize()
```

Peter9192:
changes implemented, can you approve and merge?

ruudvdent (approved):

PR #356, opened 2024-04-04
Calculate with [kg m-2] instead of [m3]

Peter9192:
Redo #349 with cleaner commits.


Peter9192:
Weird.. I thought the tests shouldn't pass.

Peter9192:
> > Weird.. I thought the tests shouldn't pass.
> 
> Well they did not run. The tests only run on PR to main:
> 
> https://github.com/WAM2layers/WAM2layers/blob/37b5b1026d93ccd0fc514ee00fb59ff998ba5f25/.github/workflows/ci.yaml#L3C1-L10C1

That explains. I didn't look into it really, had to leave and wanted to make a note to self. Thanks for clarifying in the meantime 😄

Peter9192:
Merge after #355

src/wam2layers/utils/grid.py
```diff
@@ -0,0 +1,68 @@
...
+
+def stagger_y(f):
+    """Interpolate f to the grid cell interfaces.
+    Only the values at the interior interfaces are returned
+    Arguments:
+        f: 2d array of shape [M, N]
+    Returns:
+        2d array of shape [M-1, N-2]
+    """
+    return 0.5 * (f[:-1, :] + f[1:, :])[:, 1:-1]
```
BSchilperoort, line 26: Aren't you neglecting the change in gridcell width (km) as you move north/south here? This is not a conservative regridding step. You probably have a net loss in moisture due to this step, assuming closer to the equator there is more moisture.

(Is this accounted for at all in wam2layers?)

BSchilperoort (commented):
Nice work Peter! :rocket:

I mostly focused on the technical aspects of the code, I'll leave the scientific part to @ruudvdent ;)

You did not modify `src/wam2layers/reference/variables.py`, which contains the variable references (and is still in m3). Do you intend to do that still in this PR?

src/wam2layers/utils/grid.py
ruudvdent: If this was a problem before (interpolating between m3/timestep fluxes), then it's not a problem anymore as we are now interpolating kg m-1 s-1, which is not dependent on the shape of the grid cells

src/wam2layers/tracking/core.py
```diff
@@ -123,3 +126,122 @@ def vertical_dispersion(fv, s_lower, s_upper, kvf):
     dispersion = kvf * |Fv| * dS/dz
     """
     return kvf * np.abs(fv) * (s_upper - s_lower)
+
+
+def stabilize_fluxes(F, S, progress_tracker: ProgressTracker, config: Config, t):
+    """Stabilize the outfluxes / influxes.
+
```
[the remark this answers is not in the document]
Peter9192: Done

PR #362, opened 2024-04-05
Remove snapshots

### LINKED ISSUE #322: Rework visualization/analysis module
Currently the WAM2layers has an `analysis` module, and a `visualize` command in the CLI. Here, I think we're trying to combine two purposes that don't like to be combined:

1. Quickly visualize the result of a simulation, in order to assess whether it was successful. This should be very simple and not much customizable. It should work for forward track and backward track, and it should be able to visualize both the cumulative sources and the cumulative sinks. Perhaps in the future it could also support creating a small GIF, like in #201
2. Do a more in-depth study of the outputs in an interactive (notebook) session; it would be great if common functions like loading all output data, all input data, the source region, making a nice map with cartopy, would be available as functions that can be imported from the analysis module, plot a single frame, etc. Ideally, the command line script used for 1 can import functions from the visualization module. This module should be documented with example notebooks and perhaps API docs. 

Refs #160, #208, #226, #318, #201

Peter9192:
This PR removes the cli command to generate snapshots. This was deemed to unstable for the v3 release. We can "build back better" after the v3 release.

closes #322.


BSchilperoort (approved):
Looks good. Would you want to track re-adding this in an issue, or is an issue for this already present?

Peter9192:
> Would you want to track re-adding this in an issue, or is an issue for this already present?

I think it can be added again later once we have a clearer idea of what we want/need.

Peter9192:
Thanks for the review!

Issue #360, opened 2024-04-05
ERA5 preprocessing: consistent time coordinates


BSchilperoort:
The ERA5 data used in the preprocessor has some quirks.

Accumulated variables are from the preceding hour (e.g. precipitation, evaporation).
Instantaneous variables (moisture content, wind) are "at the timestamp".

This causes a misalignment if you want to define time stamps. The rest of the code in WAM2layers also assumes that all input variables have the same time bounds.

To solve this issue:

- [ ] Homogenize time coordinates for all ERA5 variables
- [ ] Add these time bounds to the output data

ruudvdent:
I agree, this is important, not only for the output of the pre-processing data itself, but in fact it is also something ambigous what *_end_date in the config means precisely.

ruudvdent:
We agree that it would be nice to make ERA5 follow CMIP via ESMVALTOOL

ruudvdent:
Important to check how this works through in tracking. Could potentially destroy backward compatibility.

PR #363, opened 2024-04-05
Try to follow CF-convention in preprocessed and output data

### LINKED ISSUE #335: Adhere to CF-conventions in netCDF output
WAM2Layers computes accumulated variables. For these the time bounds are essential to be able to interpret the output data correctly.

> It is often the case that data values are not representative of single points in time and/or space, but rather of intervals or multidimensional cells. This convention defines a bounds attribute to specify the extent of intervals or cells. 

See http://cfconventions.org/cf-conventions/v1.6.0/cf-conventions.html#cell-boundaries

We could (try to) adhere to the CF-conventions in the output file this would look similar to:
image

_Originally posted by @BSchilperoort in https://github.com/WAM2layers/WAM2layers/pull/334#discussion_r1502256861_

BSchilperoort:
Closes #335 

Follow up in #360


src/wam2layers/preprocessing/era5.py
```diff
@@ -1,24 +1,43 @@
...
+    "title": "ERA5 data preprocessed for use in WAM2layers",
+    "history": (
+        f"created on {pydt.utcnow():%Y-%m-%dT%H:%M:%SZ} "
+        f"using wam2layers version {__version__}."
+    ),
+    "source": (
+        "ECMWF Reanalysis v5 (ERA5), "
+        "www.ecmwf.int/en/forecasts/dataset/ecmwf-reanalysis-v5"
+    ),
+    "references": "doi.org/10.5194/esd-5-471-2014",
```
Peter9192, line 36: concept doi of WAM2layer software?

Peter9192 (approved):
Nice work!

PR #368, opened 2024-04-09
Download example data

Peter9192:
This PR add a command to the wam2layers cli for downloading example data:

```sh
wam2layers download example-input-eiffel
wam2layers download example-input-volta
```

I decided to stick with a plain "download" command and leave most of the semantics to the case name. In that way, we can easily add more cases in the future. For example, we could extend it with e.g. `wam2layers download example-output-eiffel` et cetera. 

To do:

- [x] wait for publish of datasets and update download links (and check if they work):
  - [x] https://doi.org/10.4121/bbe10a2a-39dc-4098-a69f-0f677d06ecdd.v1
  - [x] https://doi.org/10.4121/f9572240-f179-4338-9e1b-82c5598529e2.v1
- [x] remove example dataset, or move it to tests
- [ ] Update documentation (see #367)
- [x] Check implementation of 4TU downloader
- [ ] Add config and region files to uploads


Peter9192:
In implementing the download functionality, I do think it might be easier to include the config file and tracking region after all. 

Advantage: the 4tu entry is a self-contained case that can be run out of the box. No need to grab files from different places.
Disadvantage: if config-file or tagging_region changes, we need update the dataset. However, that means it will be backward incompatible (breaking changes), so that should be discouraged anyway, and warrants a new version.

For the config file, the paths should be relative. @Imme1992 @ruudvdent would you still be able to add these files to the 4TU uploads?

Peter9192:
@BSchilperoort not ready yet but you could already test/review. I can make it ready once the datasets are public. I asked Stefan to have a look at my 4TU downloader as I'm not confident about the streaming extraction.

src/wam2layers/examples.py
```diff
@@ -0,0 +1,58 @@
...
+import zipfile
+from pathlib import Path
+from typing import Callable
+
+import requests
+
+
+def extract_from_4TU_archive(download_link, target_dir):
+    # See https://stackoverflow.com/a/14260592
+    response = requests.get(download_link, stream=True)  # TODO stream necessary or not?
```
sverhoeven, line 13: Reading response.content creates a bytes var of whole zipfile in memory.
To reduce memory footprint you could be use something like
```python
import requests
import zipfile
import tempfile
import shutil

url = 'http://localhost:8000/run1-full.zip'
# url = 'http://localhost:8000/docking-protein-protein-full-local.zip'
target_dir = 'bla'
with requests.get(url, stream=True) as response:
    with tempfile.TemporaryFile() as f:
        shutil.copyfileobj(response.raw, f)
        f.seek(0)
        with zipfile.ZipFile(f) as z:
            z.extractall(target_dir)
```

context on get() is needed to close connection
copyfileobj() is needed because response.raw is a file like object but cant seek, which is needed by ziplfile.

sverhoeven: Instead of TemporaryFile you could use BytesIO if file is always small or SpooledTemporaryFile if you want small file to  not be written to disk and big file should.

src/wam2layers/examples.py
```diff
@@ -0,0 +1,58 @@
...
+    print(f"Downloading and extracing random test dataset to {target_directory}")
+    extract_from_4TU_archive(download_link, target_directory)
+
+
+def download_input_volta() -> None:
+    # TODO: replace with published link
+    download_link = "https://data.4tu.nl/ndownloader/items/bbe10a2a-39dc-4098-a69f-0f677d06ecdd/versions/draft"
+    target_directory = Path.cwd() / "example_volta"
+
+    print(
```
BSchilperoort, line 36: The download is 9GB, so it might be good to warn the user that it can take some time (as there is no progress bar).

If the only way that this function is run is through the CLI, `click.echo` is recommended over `print`.

BSchilperoort:
I downloaded the 1998 Volta case, but there is no config file, so I wasn't able to run it. The West Europe 2021 data also does not have a config file.

I believe we discussed shipping the config files in the wam2layer packages, making it easier to keep them up to date with the code.

ruudvdent:
@BSchilperoort , work in progress. Ik ga er tijdens EGU met samen met Imme nog even voor zitten, dan zullen we ook de config-file etc. uploaden.

src/wam2layers/examples.py
Peter9192: Thanks! I think I get most of it now :-)
In the end I did a file-by-file download, but this will surely be useful in the future

Peter9192:
Download now uses 4TU API endpoint to get the thredds catalog then a file-by-file download. Also, it operates directly on the DOI instead of the 4TU custom links.

src/wam2layers/examples.py
Peter9192: Now it logs a message for every file, so you see the progress. Changed to logging, don't want the CLI/click to seep through to the library code.

Peter9192:
Waiting for update of 4TU datasets to include config file and region mask.

Peter9192:
Just ran the whole workflows for example-input-eiffel and this works from start to finish 🎉

Peter9192:
example-input-volta also works now, except for the last timestep in the tracking. It would be good to fix that, but this confirms that the functionality in this PR works, so I think this can be merged now.

BSchilperoort (approved):
I was able to run the cases, works OK some issues I encountered I've reported in the #367 PR

PR #379, opened 2024-05-24
Bugfix periodic boundary

### LINKED ISSUE #378: No moisture source output when putting periodic boundaries on True
I have been simulating the moisture sources for an extreme precipitation case over Australia, where I want to use the global domain.
However, when I set in my config the periodic boundaries on True I do not get any output in e_track, neither in s_track_upper_restart and s_track_lower_restart.
When I set the periodic boundaries to false in the config, I do get logic e_track values etc. 

So there seems to go something wrong when periodic boundaries are applied? I have global data on a grid from -180 to 180.

Peter9192:
Fixes #378 

I think I found the bug described in #378. With this fix, after one day of tracking, I get:

```
2022-02-27 00:00:00 - Tracked moisture: 5.50%. Tagged in atmosphere: 94.24%. 
    Boundary transport: 0.00%. Lost moisture: 0.26%. Gained moisture: 0.00%. Tagging closure: 100.00%. 
```

So still some lost moisture, but not the 100% that we saw before.

### DIFF
--- a/src/wam2layers/tracking/core.py
+++ b/src/wam2layers/tracking/core.py
@@ -101,7 +101,7 @@ def horizontal_advection(s, u, v, periodic_x=False) -> np.ndarray:
     # Donor cell upwind scheme (2 directions seperately)
     uq = np.where(up > 0, up * qp[inner, west], up * qp[inner, east])  # [M, N+1]
 
-    vq = np.where(vp > 0, vp * qp[south, inner], vp * qp[north, inner])  # [M, N+2]
+    vq = np.where(vp > 0, vp * qp[south, inner], vp * qp[north, inner])  # [M+1, N]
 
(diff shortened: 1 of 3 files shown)


Peter9192:
@Imme1992 could you try again with this branch? Feel free to run `wam2layers track config-peter.yaml` and overwrite my output if that's easier. This seems to be quite demanding in terms of compute, so would be fair to run on compute node. And we should find more funding to speed it up again :-)

Imme1992 (approved):
I tested the bug fix, it works and I do now get e_track outputted and flowing over the boundaries, thanks a lot! (while having periodic boundaries on True)

Peter9192:
Thanks, glad the issue is solved! I just realized I forgot to add a changelog entry 😞 .

Issue #207, opened 2023-04-21
Check for high altitude functionalities


ruudvdent:
### Discussed in https://github.com/WAM2layers/WAM2layers/discussions/190

<div type='discussions-op-text'>

<sup>Originally posted by **allayur** March 15, 2023</sup>
I am running current version of WAM2layers for Tibet source_region 
In preprocessing step I often get error 
assert np.all(dp > 0), "Pressure levels should increase monotonically"
AssertionError: Pressure levels should increase monotonically
I automatically download ERA5 by the script provided
Can it because the surface pressure for the high altitude is low? I run on pressure levels. The last one is 1000</div>

PR #387, opened 2024-06-17
Restore pressure layer preprocessing

### LINKED ISSUE #350: xarray==2024.3.0 breaks tests
With the newest xarray version, results from the calculations (pre-processing, backtracking) are slightly different. The difference is ~0.002% but that's enough to make the tests fail.

It also is a larger error than what I'd expect with, for example, floating point errors.

Due to the nature of the data there are a lot of 0 or near-0 values in the data we try to test for regression. This makes it difficult to easily test if the changes in the data are significant or not, as relative errors (per data point) can easily get large.

For example, on the pre-processing:
```
E           Max absolute difference: 5.820766e-10
E           Max relative difference: 0.40405387
```
The largest absolute difference is quite minor, but the max relative difference is 40%.

I guess that per variable we can define what deviation is acceptable, in appropriate units (e.g. a kg/m2/s value that is translated from a value in mm/day).

### LINKED ISSUE #376: logging info on screen for preprocessing
The logging for the tracking is now very clear, but the preprocessing logging is still a bit vague:

> 1998-07-23 22:00:00
> 1998-07-23 23:00:00
> 1998-07-24 00:00:00
> Total column water correction: mean over grid for this timestep 0.9943
> 1998-07-24 01:00:00
> 1998-07-24 02:00:00
> 1998-07-24 03:00:00

It is not immediately clear to what correction we refer. I think this is the ratio of total column water vapor to total column water, but whatever it is it should be made more clear.

BSchilperoort:
Closes #376, #207

The preprocessing functionality broke due to a change in the rest of the preprocessing code, as it depended on a certain number & order of dimensions.

- The pressure layer preprocessing code has been overhauled to only use numpy in the extension of layers.
- The two steps (1. extending the layers to atmosphere, surface and boundary, 2. interpolating to midpoints) have been split up into separate functions.
- Documentation (docstring) on the functions has also been added.
- A basic test has been added that at least tests if we can run the pressure layer preprocessing.

<hr>
Additionally:

- Tests failed due to small floating point differences. I set up some absolute/relative tolerances to avoid this. Closes #350 
- I noticed that the test verification files still had the old attributes, so in https://github.com/WAM2layers/WAM2layers/pull/356 the values were updated of the regression test files, but not the attributes. I updated them now.


tests/test_workflow.py
```diff
@@ -22,8 +22,12 @@
 
 
 CFG_FILE_BACKWARD = Path("tests/test_data/config_rhine.yaml")
+CFG_FILE_BACKWARD_PL = Path("tests/test_data/config_rhine_pl.yaml")
 CFG_FILE_FORWARD = Path("tests/test_data/config_west_africa.yaml")
 
+RTOL = 1e-5  # 0.001 %
+ATOL = 1e-7  # kg/m2/s. Approx 0.009 mm/day?
```
BSchilperoort, line 29: @Peter9192 does this seem correct?

Peter9192: I'm not sure. You use it both for preprocess (precip) and tracking (etrack/ptrack), do we expect similar values there? Also the tracking vars are accumulations since previous output, no? So no flux units.

Relative tolerance seems fine to me

src/wam2layers/preprocessing/era5.py
```diff
@@ -185,45 +193,63 @@ def get_dp_pressurelevels(q, u, v, ps, qs, us, vs):
...
-    q = insert_level(q, 0, 0)
-    p = insert_level(p, 0, 0)
-
-    # Insert surface level values (at a high dummy pressure value)
-    u = insert_level(u, us, 110000)
-    v = insert_level(v, vs, 110000)
-    q = insert_level(q, qs, 110000)
-    p = insert_level(p, ps, 110000)
+    # Assume wind at top same as values at lowest pressure, humidity at top = 0
+    p0 = _p.mean(axis=spatial_axes).argmin()  # should be a cleaner way to do this.
```
Peter9192, line 209: Why take the spatial mean?

tests/test_workflow.py
BSchilperoort: We can set different values for each. Absolute tolerance is necessary as otherwise near-0 values easily have large relative errors due to floating point inaccuracies.

src/wam2layers/preprocessing/era5.py
```diff
@@ -185,45 +193,63 @@ def get_dp_pressurelevels(q, u, v, ps, qs, us, vs):
...
+    level_ax: int = u.get_axis_num("level")  # type: ignore
+    spatial_axes = u.get_axis_num(("latitude", "longitude"))
 
     p = u["level"].broadcast_like(u) * 100  # Pa
     p.attrs["units"] = "Pa"
 
+    _p = p.to_numpy()
+    _u = u.to_numpy()
+    _v = v.to_numpy()
+    _q = q.to_numpy()
```
Peter9192, line 205: Does this create copies, i.e. does it double the memory use? If so, could temporarily delete xr arrays until you convert back on line 247

Peter9192 (commented):
Nice porting this to numpy. I had two small questions on that. Not looked at the rest of the PR

src/wam2layers/preprocessing/era5.py
```diff
@@ -304,8 +361,8 @@ def prep_experiment(config_file):
             upper_layer = ~lower_layer
 
         if config.level_type == "pressure_levels":
-            upper_layer = p < pb[:, None, :, :]
-            lower_layer = pb[:, None, :, :] < p
+            upper_layer = p < pb[None, :, :]
+            lower_layer = pb[None, :, :] < p
```
Peter9192, line 365: Would [`broadcast_like`](https://docs.xarray.dev/en/stable/generated/xarray.DataArray.broadcast_like.html) be a way to make this independent of the total dimensionality?

src/wam2layers/preprocessing/era5.py
BSchilperoort: This is the line I wanted to improve. I'm not sure what the best way of taking the slice `_p[:,0,0]` is programmatically (i.e. with the `level_ax` variable... :thinking:

src/wam2layers/preprocessing/era5.py
```diff
@@ -205,15 +260,15 @@ def get_dp_pressurelevels(q, u, v, ps, qs, us, vs):
     assert np.all(dp >= 0), "Pressure levels should increase monotonically"
 
     # Interpolate to midpoints
-    midpoints = 0.5 * (u.level.values[1:] + u.level.values[:-1])
+    midpoints = 0.5 * (u["level"].to_numpy()[1:] + u["level"].to_numpy()[:-1])
```
Peter9192, line 263: What's the difference between `.values` and `to_numpy`?

tests/test_data/era5_pl/prepare_test_data.py
```diff
@@ -0,0 +1,26 @@
+"""Prepare test data using ERA5 model level fields.
+
+This script processes daily ERA5 field on multiple model levels.
```
Peter9192, line 3:
```suggestion
This script processes daily ERA5 field on multiple pressure levels.
```

tests/test_data/era5_pl/prepare_test_data.py
```diff
@@ -0,0 +1,26 @@
...
+The data can be downloaded via CDS. More information about the
+variables and fields that are needed for tests can be found in
+our example ERA5 data downloading script in "scripts/download_era5_ml.py".
```
Peter9192, line 6:
```suggestion
our example ERA5 data downloading script in "scripts/download_era5_pl.py".
```

Peter9192 (approved):
LGTM overall.

src/wam2layers/preprocessing/era5.py
BSchilperoort: Posted this above already, but `.to_numpy()` does the same as `.values` but is more explicit. In pandas [they are already asking](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.values.html) for users to move to this syntax, in xarray not yet.

src/wam2layers/preprocessing/era5.py
Peter9192: What about `_p.argmin(axis=level_ax)`?

Peter9192: Also, I think you want the slice `_p[0, :, :]`, right?

src/wam2layers/preprocessing/era5.py
Peter9192: Thanks, didn't know that!

src/wam2layers/preprocessing/era5.py
BSchilperoort: I just need to know the index along the `level_ax` where the lowest pressure is.

```py
_p.argmin(axis=level_ax).flatten()[0]
```
Seems to work well.

```py
_p.argmin(axis=level_ax)
```
gives a 2d array, which I don't want here.

src/wam2layers/preprocessing/era5.py
BSchilperoort: Yes, but I did have to make `pb` a DataArray. But that works.

src/wam2layers/preprocessing/era5.py
BSchilperoort: With the changes to the code structure I am not sure if it's a good idea to delete the DataArrays anymore.

To reduce the memory use, we can load the DataArrays lazily (in the `load_data` function).

tests/test_workflow.py
BSchilperoort: I was looking at the test verification files, these still had the old attributes so in #356 the values were updated of the regression test files, but not the attributes. I updated them now.

All variables we are looking at have the unit of `kg m-2`. This is 1 mm water depth equivalent. So an absolute tolerance of 1e-7 mm should be good enough here.

PR #391, opened 2024-06-18
Generalize preprocessing module

BSchilperoort:
To allow for easier addition of additional datasets, the preprocessing module needs some refactoring.
Before I started the refactoring I added a regression test, to ensure results do not change.

- Instead of using loose variables (u, v, q, etc), the input data is now stored in a single xr.Dataset
  - this allows an easier definition of "what input data is needed for wam2layers"
- The more generic code has been moved to a `preprocessing/shared.py` module.
  - the old `shared.py` is now `utils.py`
  - ERA5 specific code stays in `era5.py`

Todo:
- [x] Further split up logic/data loading
- [x] Add input data checker
  - Does it have the required variables
  - Are the input dims and coords valid
  - Are the (pressure) levels monotonous
- [ ] ~~Add unit tests~~


src/wam2layers/preprocessing/shared.py
```diff
@@ -1,218 +1,164 @@
...
+            # Fluxes will be calculated based on the column water vapour
+            cw = cwv
+            log_once(
+                logger, "Total column water not available; using water vapour only"
+            )
+
+        # Integrate fluxes and states to upper and lower layer
+        if config.level_type == "model_levels":
+            # TODO: Check if this is a reasonable choice for boundary
+            boundary = 111
```
ruudvdent, line 130: This is quite specific for ERA5 only!

ruudvdent (changes requested):
Looks really good! I haven't actually run it yet, I can do that tomorrow. I highlighted some parts were we have to be a bit careful to call it generic/shared as that may be quite specific to ERA5 still. Let's discuss tomorrow.

src/wam2layers/preprocessing/pressure_levels.py
```diff
@@ -0,0 +1,139 @@
...
+            vs: Northward horizontal wind speed at the surface
+
+    Returns:
+        Dataset with the following variables (interpolated to new pressure levels):
+            p, q, u, v
+        The pressure at the division between the two layers
+    """
+    level_ax: int = input["u"].get_axis_num("level")  # type: ignore
+
+    da_p = input["level"].broadcast_like(input["u"]) * 100  # Pa
```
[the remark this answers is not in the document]
BSchilperoort: Good point, I can move this to the data loading part.

src/wam2layers/preprocessing/shared.py
BSchilperoort: Yeah, I did note that down in my comment. We can discuss this tomorrow

src/wam2layers/preprocessing/shared.py
```diff
@@ -1,218 +1,164 @@
...
-
-def repeat_upper_level(pressure_level_data, fill_value=None):
-    """Add one more level with the same value as the upper level.
-
-    A dummy value of 100 hPa is inserted for the top level pressure
-    """
-    top_level_data = pressure_level_data.isel(lev=-1).copy()
-    top_level_data["lev"] = 10000.0
+from wam2layers.config import Config
+from wam2layers.preprocessing.era5 import PREPROCESS_ATTRS, get_dp_modellevels, load_data, log_once, logger, preprocess_precip_and_evap
```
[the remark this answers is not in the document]
BSchilperoort: Eventually not, but this is a work in progress.

In the latest comment only the following are imported:
```py
from wam2layers.preprocessing.era5 import get_dp_modellevels, logger
```
The logger should be moved elsewhere, while the `get_dp_modellevels` function will probably have to change to support non-ERA5 models (it currently still loads a table from a .csv file).

tests/test_workflow.py
```diff
@@ -252,6 +252,22 @@ def test_preprocess(self, preprocessed_dir_pl: Path):
         output_path = preprocessed_dir_pl / "2020-01-01_fluxes_storages.nc"
         assert output_path.exists()
 
+        # verify outputs
+        expected_path = Path(
+            "tests/test_data/verify_output/preprocessed_data/2020-01-01_fluxes_storages.nc"
```
Peter9192, line 257: It is not clear from the filename/path that this is reference data for pressure levels. Should we make that more explicit?

src/wam2layers/preprocessing/pressure_levels.py
```diff
@@ -0,0 +1,139 @@
+from wam2layers.preprocessing.utils import interpolate, sortby_ndarray
+
+
+import numpy as np
+import xarray as xr
+
+
+def extend_pressurelevels(
+    input: xr.Dataset,
```
Peter9192: input vs surface is a bit confusing at first glance. Can we change it to e.g. 'input_3d', and 'input_surface'? Or pressure_level_variables vs surface_variables

src/wam2layers/preprocessing/pressure_levels.py
```diff
@@ -0,0 +1,139 @@
...
+    u = np.insert(u, 0, surface["us"], axis=level_ax)
+
+    # Sort arrays by pressure (ascending)
+    u = sortby_ndarray(u, p, axis=level_ax)
+    v = sortby_ndarray(v, p, axis=level_ax)
+    q = sortby_ndarray(q, p, axis=level_ax)
+    p = sortby_ndarray(p, p, axis=level_ax)
+
+    # Insert boundary level values (at a dummy pressure value)
+    p_boundary = 0.72878581 * np.array(surface["ps"]) + 7438.803223
```
Peter9192: Pass this into the function?

src/wam2layers/preprocessing/pressure_levels.py
```diff
@@ -0,0 +1,139 @@
...
+    output["v"] = xr.DataArray(v, coords, input["v"].dims, attrs=input["v"].attrs)
+    output["q"] = xr.DataArray(q, coords, input["q"].dims, attrs=input["q"].attrs)
+    output["p"] = xr.DataArray(p, coords, input["u"].dims, attrs=surface["ps"].attrs)
+    pb = xr.DataArray(p_boundary, surface["ps"].coords, surface["ps"].dims)
+
+    return output, pb
+
+
+def interp_dp_midpoints(
+    extended_data: xr.Dataset,
```
Peter9192: `extended_data` is a logical name only in the context outside the scope of the function. Rename to `data` or `pressure_level_data`?

src/wam2layers/preprocessing/pressure_levels.py
```diff
@@ -0,0 +1,139 @@
...
+):
+    """Interpolate the data to midpoints to allow for integration to two layers.
+
+    Args:
+        Dataset with the following variables:
+            q: Specific humidity at levels
+            u: Eastward horizontal wind speed at levels
+            v: Northward horizontal wind speed at levels
+            p: Air pressure at levels
+            ps: Air pressure at the surface
```
Peter9192: `ps` is a separate input argument. Also explain why it is needed? This function still does two things, also masking data below surface.

src/wam2layers/preprocessing/shared.py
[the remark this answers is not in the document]
Peter9192: I agree, eventually shared.py should be agnostic about input data source

src/wam2layers/preprocessing/pressure_levels.py
```diff
@@ -1,13 +1,12 @@
...
 import numpy as np
 import xarray as xr
 
+from wam2layers.preprocessing.utils import interpolate, sortby_ndarray
+
 
 def extend_pressurelevels(
-    input: xr.Dataset,
-    surface: xr.Dataset,
+    level_data: xr.Dataset,
```
Peter9192, line 8: Ahh, I commented on this in the previous commit, but here you already address it :-)

src/wam2layers/cli.py
```diff
@@ -164,7 +164,7 @@ def era5(ctx, config_file):
     _copy_config_yaml(config_file, log_path)
     logger.info("Welcome to WAM2layers.")
     logger.info("Starting preprocessing ERA5 data.")
-    prep_experiment(config_file)
+    prep_experiment(config_file, data_source="ERA5")
```
Peter9192, line 167: I like this syntax. Could we eventually connect this to the command line argument as well? Now we have "wam2layers preprocess era5 ...". It could either get the options from supported data sources, or we could modify the config file to include data source, and get rid of this CLI option altogether.

Peter9192 (commented):
Nice work! I reviewed commit by commit, some of my comments were already addressed in subsequent commits.

src/wam2layers/preprocessing/shared.py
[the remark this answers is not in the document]
BSchilperoort: I've made changes to make shared.py fully agnostic about input data sources

src/wam2layers/preprocessing/shared.py
BSchilperoort: Made this generic, and an optional model config

tests/test_workflow.py
BSchilperoort: That's slightly challenging: we want to make it easy to overwrite the regression data with new data. If we change the filenames, the output filenames won't be the same anymore either. So instead of just overwriting the regression test files, you have to copy the new one, delete the old one, and rename the new one.

src/wam2layers/preprocessing/pressure_levels.py
BSchilperoort: I've made it configurable, so now we only need to pass the config to the function

src/wam2layers/preprocessing/pressure_levels.py
BSchilperoort: Good point. I split off the masking to a separate function

Peter9192 (approved):
Great work!

ruudvdent (approved):

PR #392, opened 2024-06-20
Improve documentation/output

### LINKED ISSUE #381: understanding the output
Hello.
I would like to understand the output variables that are in the netcdf (example: backtrack_*.nc).
- the description of the variable "e_track" says: kg m-2 accumulated per output time step. What is the output time step? Is it the timestep that is set in the config_file.yaml file?
- the description of the "s_track_lower_restart" variable reads: tracked atmospheric moisture storage state in lower layer exactly at the out time step. Is "out time step" the same as output time step? or does it refer to the "output_frequency" set in the config_file.yaml file?
I would appreciate some clarification on this.

### LINKED ISSUE #383: Add instructions for windows users.
Why do you recommend this for novice users? There is no reason why the package wouldn't work on native windows, even cartopy has Windows (and ARM Mac) wheels now.

_Originally posted by @BSchilperoort in https://github.com/WAM2layers/WAM2layers/pull/367#discussion_r1632642249_

BSchilperoort:
Closes #383 
Closes #381


docs/userguide/installation.md
```diff
@@ -32,6 +33,10 @@ your system. If it's not,
 [here](https://docs.anaconda.com/free/miniconda/index.html) is a really nice
 walkthrough for installing miniconda.
 
+Alternatively, you can install Python by downloading it from [python.org](https://www.python.org).
```
Peter9192, line 36: Note that `which` is not available on powershell, so the above check will fail. Might want to rewrite this a little

Peter9192 (approved):
Almost good to go

ruudvdent (approved):

PR #394, opened 2024-06-21
Add CMIP data loader

BSchilperoort:
This PR took more effort than initially hoped: making all code compatible with [CF timestamps](https://github.com/Unidata/cftime) was more work than anticipated.

**Overview of changes** (in a more logical order than the commits):

- Changed input dates from `datetime.datetime` to `cftime` datetimes.
  - didn't play nice with the config's Pydantic so that code is a bit messier now :disappointed:  @Peter9192 
- All timestamps internally are now cftime based, and date ranges are `xarray.CFTimeIndex` objects
- Instead of a simple loop and [checking the hour to see if the data covers a new day](https://github.com/WAM2layers/WAM2layers/blob/ffee8b844af92a991a9cbe7d5f70a408f25498e6/src/wam2layers/preprocessing/shared.py#L94), the preprocessing dates are [split using a groupby operation](https://github.com/WAM2layers/WAM2layers/blob/ddd457dbce507ab642bf81ce707f0e63cd240296/src/wam2layers/preprocessing/shared.py#L113-L114).
  - This allowed adding an experimental parallel preprocessor (speeds up preprocessing on my machine 4-8x, so it made for much faster testing)
  - The parallel preprocessing currently only runs if explicitly enabled in the config file
- Added caching to the opening of netCDF files for a minor speed boost (using `@cache` and `@lru_cache`)
- Added CMIP preprocessor (that works with the input data listed below), you now can do `wam2layers preprocess cmip`

<hr>

**Current input data requirements**

Note that this data loader is made for the specific files from MPI which are on a 3hr and 6hr frequency: this has not been generalized yet.

If you use variables from the following tables, the preprocessor should work for other climate models as well:
- mip table "3hr" vars: `hfls`, `pr`
- mip table "6hrPlevPt" vars: `hus`, `huss`, `ua`, `uas`, `va`, `vas`
- mip table "6hrLev" var: `ps`

<details>
<summary>example config</summary>

```yaml
# General
preprocessed_data_folder: ./preprocessed_data
output_folder: ./output_data_fw
tracking_direction: forward
calendar: standard

# Preprocessing
filename_template: ./input_data/{variable}_*.nc
preprocess_start_date: "2030-03-01T00:00"
preprocess_end_date: "2030-08-01T00:00"
parallel_preprocess: True
level_type: pressure_levels
levels: [5000, 25000, 50000, 60000, 70000, 85000, 92500]

# Tracking
tracking_domain: null  # use full domain of preprocessed data
# tracking_domain: [0, 50, 10, 55]  # use subdomain of preprocessed data; west south east north
tracking_start_date: "2030-03-14T00:00"
tracking_end_date: "2030-08-01T00:00"

# Tagging
tagging_region: [-10, 35, 40, 60]  # alternative to mask file; west south east north
tagging_start_date: "2030-03-14T00:00"
tagging_end_date: "2030-05-14T00:00"

input_frequency: "6h"
timestep: 600  # timestep in seconds
output_frequency: "1D"
restart: False
periodic_boundary: true
kvf: 3
```

</details>

<hr>

**How to test/review:**

- Run download script (`scripts/download_MPI_cmip_data.sh`)
- Setup config file
- do `wam2layers preprocess cmip config_file.yaml`
  - note that it'll use the parallel preprocessor so it'll take more RAM/cpu
- do `wam2layers track config_file.yaml`

<hr>

**Todo:**

- [x] Check if results make sense (@Imme1992 ?)
- [x] Make sure output files have right calendar system
- [x] Open issues for points below;

Todo later:
- [ ] Make code more robust (fewer assumptions, i.e. input data freq check / tie vars to mip tables)
- [ ] Add regression tests
- [ ] Update documentation
  - [ ] Example CMIP config
  - [ ] CMIP example dataset
  - [ ] Which variables from which mip tables are required


BSchilperoort:
Can still be made better by adding some checks, updating attributes after they've changed, but it seems to run fine.

@Imme1992 do you want to test this? Perhaps also check that all values make sense

I ran it on my local machine. The input data is not _too_ large. I merged the scripts into a single one and put it in the download scripts folder.

src/wam2layers/preprocessing/cmip.py
```diff
@@ -0,0 +1,119 @@
+import numpy as np
+import pandas as pd
+import xarray as xr
+
+from wam2layers.config import Config
+
+WAM2CMIP = {  # wam2layers name -> cmip name
```
Peter9192, line 7: Should we adopt CMOR names as standard, as opposed to ERA5?

scripts/download_MPI_cmip_data.sh
```diff
@@ -0,0 +1,1049 @@
+#!/bin/bash
```
Peter9192, line 1: Perhaps add some documentation on how this file was obtained/generated?

Peter9192:
Would it make sense to turn this into another example case (on 4TU) and/or add regression tests for it?

src/wam2layers/preprocessing/cmip.py
BSchilperoort: We could, it seems more logical to use standard names. All names are only in the internal code anyway.
We don't exactly use ERA5 names either, so that's why I called it "wam2layers names"

I do really dislike the "evspsbl" name though.

scripts/download_MPI_cmip_data.sh
BSchilperoort: @Imme1992 can you give me the website link?

Imme1992:
Thanks, sounds great! I will test it but after my holiday (so will do it end of July/beginning of August)

Imme1992:
I am currently testing the CMIP configuration, my observations so far:
- preprocessing looks good
- output from the backtracking I have to dive in deeper
- I see the tracking files now got attributes with a title, history, source etc which is super nice. However the source: _Moisture tracking applied to ERA5 dataset:_ is not correct anymore if it is run with CMIP data

ruudvdent:
> the postprocessed evaporation data (evap in the fluxes_storages files)
You mean preprocessed?

Imme1992:
Yes preprocessed indeed, thanks for the sharp eye on friday afternoon @ruudvdent

Imme1992:
This is how we deal with negative and positive values for evaporation in ERA5 data, so I think we should apply this as well to the CMIP data?

Evaporation in ERA5 can be either positive (condensation) or negative (evaporation). In WAM2layers we consider both
precipitation and evaporation to be positive by definition. Therefore, in the case of ERA5 data, we flip the sign of evaporation
and transfer any remaining negative values to the precipitation variable instead.

ruudvdent:
In ERA5.py

```data["precip"], data["evap"] = preprocess_precip_and_evap(datetime, config)```

There is this specific function that deals with the negative values, but it seems that in the cmip.py this function is never called

```
def preprocess_precip_and_evap(
    datetime: pd.Timestamp, config: Config
) -> tuple[xr.DataArray, xr.DataArray]:
    """Load and pre-process precipitation and evaporation."""
    # All incoming units are accumulations (in m) since previous time step
    evap = load_data("e", datetime, config)
    precip = load_data("tp", datetime, config)  # total precipitation

    # Transfer negative (originally positive) values of evap to precip
    precip = np.maximum(precip, 0) + np.maximum(evap, 0)
    precip = precip

    # Change sign convention to all positive,
    evap = np.abs(np.minimum(evap, 0))

    precip = accumulation_to_flux(precip, input_frequency=config.input_frequency)
    evap = accumulation_to_flux(evap, input_frequency=config.input_frequency)
    return precip, evap
```

**Comment by BSchilperoort** [comment]
The negative evaporation is now added to the precipitation. There are no e_track values < 0 anymore.

**Comment by BSchilperoort** [comment]
I realized that this PR does not deal with non-standard calendars at all.

Calendars of climate models (in future scenarios) can often skip leap days, or use a 360-day calendar. Wam2layers currently can not handle this as it uses the standard (georgian) calendar for all time operations (with pandas).

**Comment by BSchilperoort** [comment]
f411868 moves the timestamps and date ranges in preprocessing to [cftime](https://github.com/Unidata/cftime), which supports non-standard calendars (such as a noleap or 360_day calendar).

File: src/wam2layers/preprocessing/cmip.py [hunk 3]
```diff
@@ -0,0 +1,141 @@
...
+    data["precip"] = load_cmip_var("tp", datetime, config)
+    data["evap"] = load_cmip_var("e", datetime, config)
+
+    data["precip"] = np.maximum(data["precip"], 0) + np.maximum(data["evap"], 0)
+    data["evap"] = np.abs(np.minimum(data["evap"], 0))
+
+    return data
+
+
+@cache
```
BSchilperoort, line 66: @Peter9192 as especially in CMIP data we access the same files over and over, adding this `@cache` decorator speeds up the loading significantly.

I also added it to the ERA5 loader to reduce the repeated calls to `xr.open_dataarray`, but the speed up is less significant there.

ruudvdent:
@BSchilperoort , ready for review?

BSchilperoort:
> @BSchilperoort , ready for review?

I am planning to still test the preprocessor with data from a different calendar, and move the use of cftime.datetime to the config validator (to make the code simpler). But it's almost ready to review.

I won't be able to finish everything lined out in the top post, but I will create issues to track this.

src/wam2layers/preprocessing/cmip.py
Peter9192: Nice, I used to have something similar. Just make sure to add an inline comment to explain what this does, as scientists using who might be reading/editing this code might not be familiar with it.

BSchilperoort:
@ruudvdent should be ready for review now :crossed_fingers: 

I tested a CMIP case in forward and backward tracking modes:
image
image

Outputs seem reasonable. Regression tests (ERA5) also pass.

PR #398, opened 2024-06-26
more elaborate doc in config

ruudvdent:


src/wam2layers/config.py
```diff
@@ -246,9 +254,10 @@ class Config(BaseModel):
     """
 
     input_frequency: str
-    """Frequency of the raw input data.
-
-    Used to calculated water volumes.
+    """Time frequency of the raw input data.
+    TODO: clarify if this is used during preprocessing, tracking or both, and
+    what one should do when, for example surface fluxes and humidity have
+    different frequencies.
```
ruudvdent, line 260: @BSchilperoort or @Peter9192 . This confused me a bit. I can investigate myself, but if one of you already knows the answer that might be faster

ruudvdent:
@Imme1992 , I don't want to bother Peter and Bart all the time. When you could have a quick look that would be great as well.

ruudvdent:
It's still a draft because I am awaiting feedback from Bart or Peter about `input_frequency'

src/wam2layers/config.py
Peter9192: I think it was used to convert from accumulations to fluxes. See https://github.com/WAM2layers/WAM2layers/blob/73bfd3601a4a916102cdb63e0e6dddc2f3899f24/src/wam2layers/preprocessing/utils.py#L189, which is called here https://github.com/WAM2layers/WAM2layers/blob/73bfd3601a4a916102cdb63e0e6dddc2f3899f24/src/wam2layers/preprocessing/era5.py#L116-L117

Also, it's used in reading the input data:
https://github.com/WAM2layers/WAM2layers/blob/73bfd3601a4a916102cdb63e0e6dddc2f3899f24/src/wam2layers/preprocessing/shared.py#L23 

same here:
https://github.com/WAM2layers/WAM2layers/blob/73bfd3601a4a916102cdb63e0e6dddc2f3899f24/src/wam2layers/tracking/io.py#L68-L74

ruudvdent: Thanks Peter. For now I will mention in the documentation that this settings carries over into the tracking. With other data, such as CMIP6 #394 when the timestep may even differ for different variables, we should probably work on a more robust solution.

src/wam2layers/config.py
```diff
@@ -96,17 +96,24 @@ class Config(BaseModel):
     tagging_region: Annotated[
         Union[FilePath, BoundingBox], AfterValidator(validate_region)
     ]
-    """Subdomain delimiting the source/sink regions for tagged moisture.
+    """Subdomain delimiting the tagging region from which in forward mode
+    evaporation is tagged (i.e., a source region) or from which i backward mode
```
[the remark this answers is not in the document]
ruudvdent: done

chrisweij (approved):

PR #406, opened 2024-08-16
Bugfix: sign of divergence term in residuals calculation

Peter9192:
While reviewing the manuscript I noticed an issue with the calculation of the residuals. I believe the divergence term here must be negative. The impact on the results of some test cases seems minimal, but still it'd be good to have a second pair of eyes on it.

Note: I also renamed the rather useless `foo` parameter to `R` for residuals, and reordered the calculation of `fz` to be more aligned with the manuscript.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ All notable changes to this project will be documented in this file.
 
 ## Unreleased
 
+### Fixed
+
+- A bug in the calculation of the vertical flux, where the divergence term was added instead of subtracted [#406](https://github.com/WAM2layers/WAM2layers/pull/406).
 
 ## Release v3.1.0 (2024-06-21)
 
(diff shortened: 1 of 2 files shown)


ruudvdent:
Did we not already fix a similar bug in #274?
>Verify that the derivation of Fv is correct
>Use the new state in the error correction term S_1 / S_T * err_T
Are you sure that the line of code changed here, should one-to-one correspond to the calculation in the manuscript (i.e., it might be something different: either the raw or adjusted error)

If you're sure, I'll have a more thorough look

ruudvdent (changes requested):
Great catch Peter, but let's try and verify this with the 'TODO' so we are sure it's correct now.

Peter9192:
Never mind, I was confused. In the manuscript we have the term on the RHS with a +, however, there we write $+(F_{x,i-\frac{1}{2}jk} - F_{x,i+\frac{1}{2}jk})$, i.e. convergence instead of divergence. So it *is* consistent between code and paper. I will consider changing the manuscript for more clarity.

Issue #419, opened 2024-10-21
WAM2layers fails to account for decreasing area of grid cells towards the poles


Peter9192:
@ruudvdent noticed that in the profiling code, we calculate totals without compensating for the area weights:

https://github.com/WAM2layers/WAM2layers/blob/ffee8b844af92a991a9cbe7d5f70a408f25498e6/src/wam2layers/utils/profiling.py#L70

Accounting for the weights, as per 12506f3, leads to a non-closure of the tagged moisture budget (111% at the end of the example eiffel case). This suggests that also in the rest of the code we fail to account for the decreasing area of the grid cells towards the poles.

The problem is that in calculating the zonal fluxes, we use the same length for the northern and the southern edges.

Peter9192:
The issue was introduced here: https://github.com/WAM2layers/WAM2layers/pull/356/commits/ea053ccc8cbce8033643626086b9588904bf3660.

I figured that `Fy * dx / A` could be approximated as `Fy * dx / (dy * dx) = Fy / dy`, which would retrieve the traditional advection term. First of all, this approximation is not exact, as the area of a grid cell is sort of trapezoidal rather than square. Moreover, by eliminating `dx` we remove the latitude-weighting. Later on, when we stagger the grid, this leads to both the top and the bottom interfaces getting equal weights.

BSchilperoort:
@Peter9192 isn't this fixed now?

ruudvdent:
Yes

PR #420, opened 2024-10-21
Patch for bug in meridional advection not accounting for decreasing grid cell size towards poles

Peter9192:
This PR addresses the bug reported #419. For now I've patched it by introducing `dx` and `a` in several places. I think a cleaner fix would be desirable.

Refs #419


Peter9192:
At this point I get again full budget closure for the Eiffel case. The difference in the log messages is at most 1% for all of boundary transport, tracked moisture, and tagged moisture in atmosphere. The spatial fields look slightly different as well:

Before fix:
cumulative_sources_before_fix
After fix:
cumulative_sources_after_fix

Peter9192:
For volta case, the difference is hardly noticeable:

Before fix:
cumulative_sources_before_fix

After fix:
cumulative_sources_after_fix

src/wam2layers/utils/profiling.py
```diff
@@ -67,7 +67,8 @@ def store_intermediate_states(self, output):
 
     def print_progress(self, t, output):
         """Print some useful runtime diagnostics."""
-        totals = output.sum()
+        area_weighted_output = output * np.cos(np.deg2rad(output["latitude"]))
```
ruudvdent, line 70: Is this exact or an approximation?

ruudvdent (commented):
Let's not forget the changelog :)

ruudvdent:
> At this point I get again full budget closure for the Eiffel case. The difference in the log messages is at most 1% for all of boundary transport, tracked moisture, and tagged moisture in atmosphere. The spatial fields look slightly different as well:
> 
> Before fix: cumulative_sources_before_fix After fix: cumulative_sources_after_fix



> At this point I get again full budget closure for the Eiffel case. The difference in the log messages is at most 1% for all of boundary transport, tracked moisture, and tagged moisture in atmosphere. The spatial fields look slightly different as well:
> 
> Before fix: cumulative_sources_before_fix After fix: cumulative_sources_after_fix

Before the logging nearly always reported 100.00%, so 1% unexplained is still not ideal, but maybe simply caused by the cosine approximation in the logging, whereas the actual transport may not have an issue

src/wam2layers/utils/profiling.py
Peter9192: This is exact

Peter9192:
> > At this point I get again full budget closure for the Eiffel case. The difference in the log messages is at most 1% for all of boundary transport, tracked moisture, and tagged moisture in atmosphere. The spatial fields look slightly different as well:
> 
> Before, the logging nearly always reported 100.00%, so 1% unexplained is still not ideal, but maybe simply caused by the cosine approximation in the logging, whereas the actual transport may not have an issue

It is still 100% closure, but the difference in individual terms between the previous (unweighted) and current (weighted) budgets is less typically less than 1% (e.g. boundary transport was 53% and now 54%). Sorry for unclear phrasing

ruudvdent:
> > > At this point I get again full budget closure for the Eiffel case. The difference in the log messages is at most 1% for all of boundary transport, tracked moisture, and tagged moisture in atmosphere. The spatial fields look slightly different as well:
> > 
> > 
> > Before, the logging nearly always reported 100.00%, so 1% unexplained is still not ideal, but maybe simply caused by the cosine approximation in the logging, whereas the actual transport may not have an issue
> 
> It is still 100% closure, but the difference in individual terms between the previous (unweighted) and current (weighted) budgets is less typically less than 1% (e.g. boundary transport was 53% and now 54%). Sorry for unclear phrasing

Indeed I found the same for the test cases, so that's good news!

Peter9192:
@ruudvdent I think this is ready to be merged. Can you approve?

src/wam2layers/utils/profiling.py
```diff
@@ -56,8 +58,7 @@ def store_intermediate_states(self, output):
             self.tracked += totals["p_track_lower"] + totals["p_track_upper"]
             self.total_tagged_moisture += totals["tagged_evap"]
 
-        # Don't include boundary losses in log messages
-        interior_losses = output.losses[1:-1, 1:-1].sum()
+        interior_losses = area_weighted_output.losses[1:-1, 1:-1].sum()
```
ruudvdent, line 61: For periodic boundaries this is not completely correct. Maybe not solve it now, but just add a 'TODO' or note about it at least?

src/wam2layers/utils/profiling.py
```diff
@@ -78,14 +80,15 @@ def print_progress(self, t, output):
             totals["s_track_upper_restart"] + totals["s_track_lower_restart"]
         )
 
-        # Don't include boundary losses in log messages
-        interior_losses = output.losses[1:-1, 1:-1].sum()
+        interior_losses = area_weighted_output.losses[1:-1, 1:-1].sum()
```
ruudvdent, line 83: Same comment as above, because it's a duplication as already mentioned in line 90:

For periodic boundaries this is not completely correct. Maybe not solve it now, but just add a 'TODO' or note about it at least?

ruudvdent (approved):
Please leave a note or TODO at the interior losses line that it doesn't work exactly for periodic boundaries. Everything else looks good.

Peter9192:
> For periodic boundaries this is not completely correct. Maybe not solve it now, but just add a 'TODO' or note about it at least?

Well spotted! This might also help to explain the unexpected losses and non-closure of the water balance seen in Hilde's case (refs #424). I thought I might as well fix it right away. Can you confirm this is sound?

ruudvdent (approved):
Super!

PR #426, opened 2024-12-09
In simulations with periodic boundaries, stagger the flux to the exterior edges as well

### LINKED ISSUE #424: Padding fluxes in advection calculation is a bad idea
Currently in the advection calculation we pad the boundaries of the input data. 

https://github.com/WAM2layers/WAM2layers/blob/b423a94323b0c535f2751d32f32a63982111f195/src/wam2layers/tracking/core.py#L91-L92

This works well for the state variables that exist at the grid cell center (i.e. we're adding ghost cells). However, it doesn't really make sense for the fluxes, which exist at the grid cell interfaces. After [staggering the grid](https://github.com/WAM2layers/WAM2layers/blob/b423a94323b0c535f2751d32f32a63982111f195/src/wam2layers/tracking/backtrack.py#L39-L42) the fluxes are only known at the interior interfaces, so we need a way to deal with that.

For periodic boundaries, we can interpolate the eastmost and westmost values to get the exterior interface. For non-periodic boundaries, the "old" method was to discard the outer band of grid cells. 

In fixing this, we need to make sure that we can still account the boundary losses, since we want to know how much moisture is lost along the boundaries of the domain.

Peter9192:
In calculating the horizontal advection term, padding is added to boundaries.
For periodic boundaries, this led to erroneous behaviour, since the padded Fx values would correspond to the interior rather than exterior boundary. 

This fix changes the code as follows:

If boundaries are periodic:
- Fx staggering now includes the outer edge, which is interpolated between the first and last grid cells. That value is inserted both at the right and and the left end.
  Return shape is now [M-2, N-1] or [M-2, N+1] when periodic_x = True
- Fy staggering only on interior, but doesn't discard the outermost x-values
  Return shape is now [M-1, N-2] or [M-1, N] when periodic_x = True
- In the advection calculation, padding is applied differently when domain is periodic:
  - For u and v, only the southern and northern edges are padded.
  - For s, the behaviour is same as before: in the x-direction we pad with mode=wrap, i.e. leftmost value is inserted at the right and vice versa.

Should fix #424 
Might be related to https://github.com/WAM2layers/WAM2layers/discussions/225 as well

### DIFF
--- a/src/wam2layers/tracking/backtrack.py
+++ b/src/wam2layers/tracking/backtrack.py
@@ -75,6 +75,12 @@ def backtrack(
         + tagging_mask * precip * s_upper / (s_upper + s_lower)
     )
 
+    # Remove boundary transport from s_track and bookkeep losses
+    boundary = get_boundary(losses, periodic=config.periodic_boundary)
+    losses = xr.where(boundary, (s_track_upper + s_track_lower), 0)
+    s_track_upper = xr.where(boundary, 0, s_track_upper)
+    s_track_lower = xr.where(boundary, 0, s_track_lower)
+
     # lower and upper: redistribute unaccounted water that is otherwise lost from the sytem
     overshoot_lower = np.maximum(0, s_track_lower - S0["s_lower"])
     overshoot_upper = np.maximum(0, s_track_upper - S0["s_upper"])
@@ -91,15 +97,9 @@ def backtrack(
     # At this point any of the storages could still be overfull, thus stabilize and assigns losses
     losses_lower = np.maximum(0, s_track_lower - S0["s_lower"])
     losses_upper = np.maximum(0, s_track_upper - S0["s_upper"])
-    losses = losses_lower + losses_upper
+    losses += losses_lower + losses_upper
     s_track_lower = np.minimum(s_track_lower, S0["s_lower"])
     s_track_upper = np.minimum(s_track_upper, S0["s_upper"])
-
-    # Bookkeep boundary transport as "lost moisture at grid edges"
-    boundary = get_boundary(losses, periodic=config.periodic_boundary)
-    losses += xr.where(boundary, (s_track_upper + s_track_lower), 0)
-    s_track_upper = xr.where(boundary, 0, s_track_upper)
-    s_track_lower = xr.where(boundary, 0, s_track_lower)
 
     # Update output fields
     output["tagged_precip"] += tagging_mask * precip * config.timestep
--- a/src/wam2layers/tracking/backtrack.py
+++ b/src/wam2layers/tracking/backtrack.py
@@ -76,7 +76,7 @@ def backtrack(
     )
 
     # Remove boundary transport from s_track and bookkeep losses
-    boundary = get_boundary(losses, periodic=config.periodic_boundary)
+    boundary = get_boundary(s_track_upper, periodic=config.periodic_boundary)
     losses = xr.where(boundary, (s_track_upper + s_track_lower), 0)
     s_track_upper = xr.where(boundary, 0, s_track_upper)
     s_track_lower = xr.where(boundary, 0, s_track_lower)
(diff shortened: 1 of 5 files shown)


Peter9192:
Hi @hildekoning I think this should fix the problem we saw in your simulations with periodic boundaries. Would you be able to test this? It would require checking out this branch in your clone of the repository. Let me know if you need help with that.

Peter9192:
@ruudvdent another pair of eyes on this would be good I think. Tests still pass, so the fix doesn't affect previous test cases, but we don't have a testcase for periodic boundaries.

hildekoning:
Hi Peter, very nice that you found a possible solution already! Of course I would like to test this with my data. However, I am not sure how I can use two different versions of WAM2layers on my account, as now I have the branche 'add CMIP' which might start interfering with another folder containing the improved version of WAM2layers code. Is there a way to switch between two branches, or should I finish my analysis with the 'unimproved' code, and then switch to this version?

Peter9192:
> as now I have the branche 'add CMIP' which might start interfering with another folder containing the improved version of WAM2layers code

The add CMIP branch was merged last Friday, so all changes in that branch are also contained within this branch. 

> Is there a way to switch between two branches

Yes, there is the `git checkout` command for that. See https://git-scm.com/docs/git-checkout. Normally this should allow you to easily switch between branches, unless you have local changes that have not been committed to any branch. If you do, git checkout will not work until you discard your local changes (if they're irrelevant) or commit or stash them (if you want to keep them)

> or should I finish my analysis with the 'unimproved' code

If you think you can finish your analysis properly without the fix, then you should probably focus on that first. But it would be great if you could test it at some point 😃

src/wam2layers/tracking/backtrack.py
```diff
@@ -91,16 +97,10 @@ def backtrack(
     # At this point any of the storages could still be overfull, thus stabilize and assigns losses
```
ruudvdent, line 97: At this point the storages could still be overfull _internally_, thus stabilize and assign _internal_ losses

src/wam2layers/tracking/core.py
```diff
@@ -5,27 +5,12 @@
...
-    zero_xy = partial(np.pad, pad_width=1, mode="constant", constant_values=0)
-    periodic_x = partial(np.pad, pad_width=((0, 0), (1, 1)), mode="wrap")
-    zero_y = partial(
-        np.pad, pad_width=((1, 1), (0, 0)), mode="constant", constant_values=0
-    )
-
-    if periodic:
-        return [periodic_x(zero_y(arg)) for arg in args]
-    return [zero_xy(arg) for arg in args]
+# Functions to pad poundaries of arrays
```
ruudvdent, line 8: _b_ oundaries

src/wam2layers/tracking/core.py
```diff
@@ -5,27 +5,12 @@
...
+pad_x_wrap = partial(np.pad, pad_width=((0, 0), (1, 1)), mode="wrap")
+pad_y_zero = partial(
+    np.pad, pad_width=((1, 1), (0, 0)), mode="constant", constant_values=0
+)
+pad_xy_zero = partial(np.pad, pad_width=1, mode="constant", constant_values=0)
 
 
 def horizontal_advection(s, u, v, periodic_x=False) -> np.ndarray:
```
ruudvdent, line 16: would be good to add a bit more inline comments in this function as these things are not straightforward of why what is happening, which would be very helpful if somebody revisits this code in a couple of years.

src/wam2layers/tracking/forwardtrack.py
```diff
@@ -92,16 +98,10 @@ def forwardtrack(
     # at this point any of the storages could still be overfull, thus stabilize and assigns losses:
```
ruudvdent, line 98: At this point the storages could still be overfull _internally_, thus stabilize and assign _internal_ losses

ruudvdent (approved):
Super! I have just a few suggestions to clarify the code with inline comments.

src/wam2layers/tracking/core.py
Peter9192: Done

Qpekin:
I believe this padding approach leads to a significant issue: When the domain is set to periodic, the different padding strategies applied to different variables (u, v, and s) result in inconsistent computational grid shapes. This shape inconsistency could potentially cause problems in subsequent calculations.

I've encountered this type of problem while running my data - how can this be resolved?
wam2layers

PR #489, opened 2025-09-05
3.9 -> 3.13

### LINKED ISSUE #482: Drop support for python 3.9 (and 3.10), update to python 3.13 in installation example
According to dev docs, we follow https://numpy.org/neps/nep-0029-deprecation_policy.html

Alternatively could follow https://endoflife.date/python security (bit more lenient)

In any case we can (almost) drop support for 3.9, which means we can use optional types with `| None` rather than `Typing.Optional` and use `|` for union types

ruudvdent:
closes #482


ruudvdent:
> https://github.com/WAM2layers/WAM2layers/blob/8127c92aa48c44031ec2b801b5b073a53fbfa9c5/pyproject.toml#L26C1-L26C33
> 
> I think you also need to update this one

Done

PR #490, opened 2025-09-05
Fix frequency correction

### LINKED ISSUE #485: `round_cftime` incorrectly changes some frequencies to lowercase
https://github.com/WAM2layers/WAM2layers/blob/8127c92aa48c44031ec2b801b5b073a53fbfa9c5/src/wam2layers/utils/calendar.py#L84-L99

changes some frequencies to lowercase to provide backwards compatibility wrt a deprecation in pandas (`S` for seconds now should be `s`). But there are more frequencies like `MS` that will also get lowercase now unintentially. This was recently raised in #466

Peter9192:
fixes #485 

- Use regex to find all deprecated frequency strings and update them to the new convention
- Immediately fix frequencies as part of parsing the config (using [pydantic field validators](https://docs.pydantic.dev/latest/concepts/validators/#using-the-annotated-pattern))
- Add tests to ensure the frequency fix works as expected

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ All notable changes to this project will be documented in this file.
 
 - updated reference to WAM2layers v3 paper
 
+### Fixed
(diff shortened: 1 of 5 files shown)


ruudvdent (approved):
awesome!

